PR #151, opened 2020-05-26
AAU, on playground, when I look at function assets i see a field created_at

p-desaintchamas:


p-desaintchamas:
Add the field label_created_at in the assets function
Fix the fields argument in function export_labels_as_df

PR #223, opened 2020-09-25
AAU, on playground, when I use webhooks, I understand why they are slow

PaulKili:
- [x] I opened a merge request on `kili` on a branch with the same name than the branch of the current pull request


kili/subscriptions/label/__init__.py
```diff
@@ -31,10 +31,13 @@ def label_created_or_updated(self, project_id: str, callback: Callable[[str, str
         - None
         """
         ws_endpoint = self.auth.client.endpoint.replace('http', 'ws')
+        # if 'v1' in ws_endpoint:
```
marcenacp, line 34: Remove those comments

marcenacp (approved):

PR #229, opened 2020-10-09
AAU VitadX, when I modify assets, I can do it by batch

mDuval1:
- [x] I opened a merge request on `kili` on a branch with the same name than the branch of the current pull request

### DIFF
--- a/kili/helpers.py
+++ b/kili/helpers.py
@@ -158,8 +158,5 @@ def convert_to_list_of_none(array, length):
 
 
 def is_not_none_or_empty(object):
-    if object is None:
-        return False
-    if isinstance(object, list) and len(object) == 0:
-        return False
-    return True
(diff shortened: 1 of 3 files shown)


kili/helpers.py
```diff
@@ -133,3 +133,33 @@ def wrapper(*args, **kwargs):
...
+        if len(array) != length:
+            raise Exception(f'array should have length {length}')
+        else:
+            return array
+    else:
+        return [None] * length
+
+
+def is_not_none_or_empty(object):
+    if object is None:
```
marcenacp, line 161:
```
object_is_empty = isinstance(object, list) and len(object) == 0
return object is not None and not (object_is_empty)
```

mDuval1: Done

PR #314, opened 2021-03-30
AAU, on playground, when I query assets, I see I can query the assets metadata

p-desaintchamas:
- [x] I opened a merge request on `kili` on a branch with the same name than the branch of the current pull request and forced security tests

### DIFF
--- a/kili/helpers.py
+++ b/kili/helpers.py
@@ -90,16 +90,12 @@ def format_json(result):
     if isinstance(result, dict):
         for key, value in result.items():
             if key in ['jsonInterface', 'jsonMetadata', 'jsonResponse']:
-                if key == 'jsonMetadata' and is_url(result[key]):
-                    print('escape')
-                    result[key] = result[key]
-                elif value == '' or value is None:
+                if (value == '' or value is None) and not (is_url(value) and key == 'jsonInterface'):
                     result[key] = dict()
                 else:
                     try:
                         result[key] = loads(value)
                     except:
-                        print(key, value)
                         raise ValueError(
                             'Json Metadata / json response / json interface should be valid jsons')
             else:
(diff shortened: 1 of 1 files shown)


kili/helpers.py
```diff
@@ -90,12 +90,16 @@ def format_json(result):
     if isinstance(result, dict):
         for key, value in result.items():
             if key in ['jsonInterface', 'jsonMetadata', 'jsonResponse']:
-                if value == '' or value is None:
+                if key == 'jsonMetadata' and is_url(result[key]):
+                    print('escape')
+                    result[key] = result[key]
```
marcenacp, line 95: First part of the `if` could be skipped?

kili/helpers.py
```diff
@@ -90,12 +90,16 @@ def format_json(result):
     if isinstance(result, dict):
         for key, value in result.items():
             if key in ['jsonInterface', 'jsonMetadata', 'jsonResponse']:
-                if value == '' or value is None:
+                if key == 'jsonMetadata' and is_url(result[key]):
+                    print('escape')
```
marcenacp, line 94: Print

kili/helpers.py
```diff
@@ -90,12 +90,16 @@ def format_json(result):
...
+                    result[key] = result[key]
+                elif value == '' or value is None:
                     result[key] = dict()
                 else:
                     try:
                         result[key] = loads(value)
                     except:
+                        print(key, value)
```
marcenacp, line 102: Print

kili/helpers.py
p-desaintchamas: What do you mean ? We can do only the check is_url(value) ?

marcenacp (approved):

PR #379, opened 2021-06-29
Script to train Resnet50 with noise and reduced dataset

FrankFacundo:


recipes/data-quality/Train_Resnet50_noise_and_reduced.py
```diff
@@ -0,0 +1,169 @@
+# coding: utf-8
```
marcenacp, line 1: - The script is not packaged enough (for instance, where should the data be stored? etc)
- The script is not clean enough to be added to `master` (comments and code clarity)

PR #422, opened 2021-09-09
AADS, when I request assets of dataset assets, I see a success

kili-technology:
- [ ] I opened a merge request on `kili` on a branch with the same name than the branch of the current pull request and forced security tests

### DIFF
--- a/kili/queries/asset/__init__.py
+++ b/kili/queries/asset/__init__.py
@@ -37,7 +37,7 @@ class QueriesAsset:
                disable_tqdm: bool = False,
                external_id_contains: Optional[List[str]] = None,
                first: Optional[int] = None,
-               format: Optional[str] = None, 
+               format: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


kili/queries/asset/__init__.py
```diff
@@ -322,6 +327,9 @@ def count_assets(self, asset_id: Optional[str] = None,
                 },
                 'updatedAtGte': updated_at_gte,
                 'updatedAtLte': updated_at_lte,
+                'datasetAsset': {
```
p-desaintchamas, line 330: @mDuval1 is it a breaking change ? can you test it with the previous version of the playground ?

baptiste-olivier (approved):

kili/queries/asset/__init__.py
mDuval1: No, not breaking

p-desaintchamas (approved):

PR #456, opened 2021-11-03
AAU, on Kili, when I create an API key, I see it expires after 1 year

mDuval1:
- [x] I opened a merge request on `kili` on a branch with the same name than the branch of the current pull request and forced security tests

### DIFF
--- a/kili/types.py
+++ b/kili/types.py
@@ -134,12 +134,31 @@ class ProjectUser(ProjectUserWithoutProject):
 
 
 @dataclass
-class User(UserWithoutProjectUsers):
+class UserWithoutApiKey(UserWithoutProjectUsers):
     """
     A wrapper for User GraphQL object.
     """
     projectUsers = ProjectUser
 
+@dataclass
+class ApiKey:
+    """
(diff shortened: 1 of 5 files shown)


kili/types.py
```diff
@@ -134,12 +134,31 @@ class ProjectUser(ProjectUserWithoutProject):
...
+    revoked = 'revoked'
+    user = UserWithoutApiKey
+    userId = 'userId'
+
+@dataclass
+class User(UserWithoutApiKey):
+    """
+    A wrapper for User GraphQL object.
+    """
+    apiKeys = ApiKey
```
p-desaintchamas, line 161: apiKeys: list[ApiKey] would be better No ? It seems it is not handled in other types but it can be done https://docs.python.org/3/library/dataclasses.html

p-desaintchamas: Ok, it's how we build the function, I create a technical debt on this issue

p-desaintchamas (approved):

marcenacp (approved):

PR #489, opened 2022-01-12
Fix delete project and make project public in the playground

hugo-mailfait-kili:
- [ ] I opened a merge request on `kili` on a branch with the same name than the branch of the current pull request and forced security tests

### DIFF
--- a/kili/mutations/project/queries.py
+++ b/kili/mutations/project/queries.py
@@ -2,7 +2,7 @@
 Queries of project mutations
 """
 
-from .fragments import PROJECT_FRAGMENT, PROJECT_FRAGMENT_ID, ROLE_FRAGMENT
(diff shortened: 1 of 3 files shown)


baptiste-olivier (approved):

kili/mutations/project/queries.py
```diff
@@ -154,10 +150,7 @@
...
 
 GQL_PROJECT_DELETE_ASYNCHRONOUSLY = '''
-mutation($projectID: ID!) {{
-  data: deleteProjectAsynchronously(where: {{
-      id: $projectID
-    }}
-    )
-}}
+mutation($where: ProjectWhere!) {
+  data: deleteProjectAsynchronously(where: $where)
```
p-desaintchamas, line 154: Did you tested this fragment ? Not sure if it would work, I think you need to return something (+ retrieving the token is useful for people turning a project public

p-desaintchamas (approved):

PR #523, opened 2022-03-21
feat(COR-1354): add check of mime type on uploading assets to project

gheorghetutunaru:

### DIFF
--- a/kili/mutations/asset/helpers.py
+++ b/kili/mutations/asset/helpers.py
@@ -7,19 +7,22 @@ from uuid import uuid4
 from typing import List, Union
 import mimetypes
 
-from ...helpers import encode_base64, is_url
+from ...constants import mime_extensions_for_IV2
+from ...helpers import encode_base64, get_data_type, is_url
 from .queries import (GQL_APPEND_MANY_TO_DATASET,
                       GQL_APPEND_MANY_FRAMES_TO_DATASET)
 
 
-def encode_object_if_not_url(content):
+def encode_object_if_not_url(content, input_type):
     """
     Return the object if it is a url, else it should be a path to a file.
     In that case, the file is returned as a base64 string
     """
     if is_url(content):
         return content
-    return encode_base64(content)
+    if check_file_mime_type(content, input_type):
+        return encode_base64(content)
+    return None
 
 
 def process_frame_json_content(json_content):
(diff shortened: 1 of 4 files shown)


kili/mutations/asset/helpers.py
```diff
@@ -74,12 +77,29 @@ def process_content(input_type: str,
...
     if input_type == 'FRAME' and json_content_array is None:
-        content_array = list(map(encode_object_if_not_url, content_array))
+        content_array = [encode_object_if_not_url(content, input_type) for content in content_array]
     return content_array
 
+def check_file_mime_type(content: str, input_type: str):
+    """
+    Returns true if the mime type of the file corresponds to the allowed mime types of the project
+    """
+    if input_type in ['IMAGE', 'FRAME', 'PDF']:
```
mDuval1, line 90: early return

```
if input_type not in ['IMAGE', 'FRAME', 'PDF']:
  return True
[...]
```

kili/mutations/asset/helpers.py
```diff
@@ -74,12 +77,29 @@ def process_content(input_type: str,
...
+        mime_type = get_data_type(content.lower())
+        if mime_extensions_for_IV2[input_type] and mime_type:
+            correct_mime_type = mime_type in mime_extensions_for_IV2[input_type]
+            if not correct_mime_type:
+                print(f'File mime type for {content} is {mime_type} and does not correspond' \
+                    'to the type of the project. '\
+                    f'File mime type should be one of {mime_extensions_for_IV2[input_type]}')
+            return correct_mime_type
+        if not (mime_extensions_for_IV2[input_type] and mime_type):
```
mDuval1, line 99: The if is not necessary, you can return False directly

kili/mutations/asset/helpers.py
```diff
@@ -74,12 +77,29 @@ def process_content(input_type: str,
...
             if (json_content_array is not None and json_content_array[i] is not None)
-                else encode_base64(content))
+            else (encode_base64(content) if check_file_mime_type(content, input_type) else None))
             for i, content in enumerate(content_array)]
     if input_type == 'FRAME' and json_content_array is None:
-        content_array = list(map(encode_object_if_not_url, content_array))
+        content_array = [encode_object_if_not_url(content, input_type) for content in content_array]
     return content_array
 
+def check_file_mime_type(content: str, input_type: str):
```
mDuval1, line 86: missing type return

kili/mutations/asset/helpers.py
gheorghetutunaru: Also added as an early return

pmKili (approved):

mDuval1 (approved):

PR #527, opened 2022-04-01
feat(ML-27): handle paginated queries inside .assets method

PierreLeveau:
**Overview**
This PR handles the asset pagination through a generator.  Additionally, when `as_generator` is passed as an argument, it outputs a generator of the asset samples. 
It also fixes the following issue: when the `first` argument was passed, it was not using pagination internally and was calling the API directly with the `first` parameter (making it crash if too high)

**Code changes**
* changed `.assets` method in `kili.queries.asset`
* added a `row_generator_from_paginated_calls` which can be used across the board to paginate the other row types.

**Test**
Added a unit test for the `.assets` method by mocking the calls to the API. Also tested locally by calling the API.

**Side effects**
May break the usage of assets everywhere, low risk since it has been tested.

### DIFF
--- a/kili/utils.py
+++ b/kili/utils.py
@@ -1,0 +1,69 @@
+"""
+Utils
+"""
+from typing import List, Callable
+
+from tqdm import tqdm
+
+# pylint: disable=too-many-arguments,too-many-locals
+def row_generator_from_paginated_calls(
+        skip: int,
+        first: int,
+        count_method: Callable[..., int],
+        count_kwargs: dict,
+        paged_call_method: Callable[..., List[dict]],
+        paged_call_payload: dict,
+        fields: List[str],
+        disable_tqdm: bool):
+    """
+    Builds a row generator from paginated calls.
+
+    Parameters
+    ----------
+    - skip : int
+        Number of assets to skip (they are ordered by their date of creation, first to last).
+    - first : int
+        Maximum number of assets to return.
+    - count_method: ... -> int
+        Callable returning the number of available assets given `count_args`.
+    - count_kwargs: dict
+        Keyword arguments passed to the `count_method`.
+    - paged_call_method: ... -> List[dict]
+        Callable returning the list of samples.
+    - paged_call_payload: dict
+        Payload for the GraphQL query.
+    - fields: List[str]
+        The list of strings to retrieved.
+    - disable_tqdm: bool
+        If `True`, disables tqdm.
+    """
+    count_rows_retrieved = 0
+    count_rows_available = count_method(**count_kwargs)
+    if not disable_tqdm:
+        count_rows_queried_total = min(count_rows_available,
+                first) if first is not None else count_rows_available
+        if count_rows_queried_total == 0:
+            yield from ()
+    else:
+        # dummy value that won't have any impact since tqdm is disabled
+        count_rows_queried_total = 1
+    count_rows_query_default = min(100, first or 100)
+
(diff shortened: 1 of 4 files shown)


test/queries/test_asset.py
```diff
@@ -0,0 +1,84 @@
...
+import os
+from kili.client import Kili
+
+api_key = os.getenv("KILI_USER_API_KEY")
+api_endpoint = os.getenv("KILI_API_ENDPOINT")
+kili = Kili(api_key=api_key, api_endpoint=api_endpoint)
+
+TEST_CASES = [
+    {
+        "case": "AAU, When I query assets with first=50, I get the first 50 assets",
```
mDuval1, line 13: I feel like a test is missing, when fetching as a generator without specifying first, to test if there aren't any problems fetching all assets

test/queries/test_asset.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    return [{"id": i} for i in range(skip, skip + count_sample)]
+
+
+def test_assets(mocker):
+    """
+        Tests several queries
+    """
+    count_sample = 1000
+    mocker.patch("kili.queries.asset.QueriesAsset._query_assets", side_effect=mocked_query_assets)
```
theodu, line 80: why side_effect here and return_value in the line below?

kili/utils.py
```diff
@@ -0,0 +1,71 @@
...
+    ----------
+    - skip : int
+        Number of assets to skip (they are ordered by their date of creation, first to last).
+    - first : int
+        Maximum number of assets to return.
+    - count_method: ... -> int
+        Callable returning the number of available assets given `count_args`.
+    - count_kwargs: dict
+        Keyword arguments passed to the `count_method`.
+    - query_method: ... -> List[dict]
```
theodu, line 31: called `paged_call_method` above

kili/utils.py
```diff
@@ -0,0 +1,71 @@
...
+        Callable returning the list of samples.
+    - payload_query: dict
```
theodu, line 33: called `paged_call_payload` above

kili/utils.py
```diff
@@ -0,0 +1,71 @@
...
+        Payload for the GraphQL query.
+    - fields: List[str]
+        The list of strings to retrieved.
+    - disable_tqdm: bool
+        If `True`, disables tqdm.
+    """
+    count_rows_retrieved = 0
+    count_rows_available = count_method(**count_kwargs)
```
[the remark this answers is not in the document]
PierreLeveau: That's not the goal for now. We want to have a generic method to handle the paginated results, so we can apply it to all the types (labels, projects, ... ). Compared to the previous implementation of `.assets`:
* it's able to provide a generator. It is interesting when there is a huge number of assets and the memory is limited, or in streaming cases.
* it handles any number of requested assets (the previous one was limited by the backend limit).
Maybe I misunderstood your comment, happy to discuss if needed 😀

test/queries/test_asset.py
PierreLeveau: `side _effect` enables to replace a method by another, `return_value` is sufficient to replace the return value.

kili/utils.py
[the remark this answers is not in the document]
PierreLeveau: counting may take time, so I removed it in the last update.

mDuval1 (approved):

edarchimbaud (approved):

PR #534, opened 2022-04-13
AADS, on Playground, when I fetch any object list, there is no limitation and can be output as a generator

theodu:

### DIFF
--- a/kili/queries/project/__init__.py
+++ b/kili/queries/project/__init__.py
@@ -2,7 +2,7 @@
 Project queries
 """
 
-from typing import List, Optional
+from typing import Generator, List, Optional, Union
 import warnings
 from typeguard import typechecked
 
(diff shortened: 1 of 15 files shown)


kili/queries/project/__init__.py
```diff
@@ -2,14 +2,19 @@
 Project queries
 """
 
-from typing import Optional
+from typing import List, Optional
+import warnings
 from typeguard import typechecked
 
+from kili.utils import row_generator_from_paginated_calls
```
PierreLeveau, line 9: -> `.utils`

kili/queries/project/__init__.py
```diff
@@ -2,14 +2,19 @@
...
+
 from ...helpers import Compatible, deprecate, format_result, fragment_builder
 from .queries import gql_projects, GQL_PROJECTS_COUNT
 from ...types import Project
 from ...constants import NO_ACCESS_RIGHT
 
+# warnings.filterwarnings("default", module='kili', category=DeprecationWarning)
```
PierreLeveau, line 16: rm comment

PierreLeveau (changes requested):

PierreLeveau (approved):

PierreLeveau:
note: I think there is an avenue for way more code factorization, but let's leave it like that for now.

p-desaintchamas (approved):

PR #540, opened 2022-04-25
AAKD, on the Python client doc, when I look at all the docs, I see an accurate representation of the Python client code

theodu:


theodu:
This is more than a POC, this is a final version

docs/README.md
```diff
@@ -0,0 +1,65 @@
+# KILI SDK docs
```
PierreLeveau, line 1: this should be part of the internal documentation in Notion right ?

docs/index.md
```diff
@@ -0,0 +1,41 @@
...
+# You can now use the Kili client!
+```
+
+Note that you can also pass the API key as an argument of the `Kili` initialization:
+
+```python
+kili = Kili(api_key='MY API KEY')
+```
+
+Here is a sample of the operations you can do with the Kili client:
```
PierreLeveau, line 41: missing end of the document?

docs/README.md
```diff
@@ -0,0 +1,65 @@
...
+## List all versions
+
+```
+mike list
+```
+
+## create a new version
+
+```
+mike deploy [version: X.XX] [alias]...
```
PierreLeveau, line 29: which Github rights does it require? (this should be available to a restricted set of internal users only)

PierreLeveau (changes requested):

docs/index.md
```diff
@@ -0,0 +1,41 @@
+# Kili's Python SDK
```
PierreLeveau, line 1: `Kili Python SDK`

mkdocs.yml
```diff
@@ -0,0 +1,66 @@
+## Project information
+site_name: Python SDK
+site_description: Kili Python SDK
+site_url: https://kili-technology.github.io/kili-playground/
+
+## Repository
+repo_url: https://github.com/kili-technology/kili-playground
+repo_name: SDK on GitHub
```
PierreLeveau, line 8: Kili Python SDK

mkdocs.yml
```diff
@@ -0,0 +1,66 @@
+## Project information
+site_name: Python SDK
```
PierreLeveau, line 2: Kili Python SDK

PierreLeveau (changes requested):

docs/README.md
theodu: Oh yeah, this could be. I saw a lot of repositories having such a readme so I put the instructions here but you are right, we don't want people to feel like they can play with it.

docs/index.md
theodu: After the discussion that we had, I wanted to put less information on the home page, just to have a website for the API reference. You can maybe think about what should be on the home page

docs/README.md
theodu: This will only make changes locally.
if a user wants to push the changes, it can add the --push argument. This will push the changes to a `gh-pages` branch and it is from here that the Github Pages will build the website. I guess that the repo does not allow external people to push to a specific branch ?

mkdocs.yml
theodu: for that one, I removed the 'Kili' in order not to see 'Kili Kili python SDK' as we have the logo just next to it...
Capture d’écran 2022-04-27 à 10 19 05

PierreLeveau: indeed.

mkdocs.yml
PierreLeveau: for this one I persist, since we have:
image

theodu: yes, I changed it in local. I am just waiting for the URL to be changed to push all the changes

docs/index.md
PierreLeveau: in this case remove the last sentence and we will be good.

theodu: ah yes, I didn't see that passing!

docs/README.md
PierreLeveau: no they can't, so we should be good.

PierreLeveau: yes I do validate.

PierreLeveau (approved):

edarchimbaud (approved):

PR #543, opened 2022-05-03
AADS, on Python SDK doc, when I read the doc, I see polished content

theodu:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -6,13 +6,7 @@ Kili is a platform that empowers a data-centric approach to Machine Learning thr
 
 ## Requirements
 
-- `python >= 3.7`
-- Create and copy a Kili API key
-- Add the `KILI_API_KEY` variable in your bash environment (or in the settings of your favorite IDE) by pasting the API key value you copied above:
-
-```bash
-export KILI_API_KEY='<you api key value here>'
-```
+You only need Python 3.7 or greater
 
 ## Installation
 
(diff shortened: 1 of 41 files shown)


kili/queries/api_key/__init__.py
```diff
@@ -128,29 +117,22 @@ def _query_api_keys(self,
...
+            api_key_id: Identifier of the API key to retrieve.
+            user_id: Identifier of the user.
+            api_key: Value of the api key.
+
+        Returns:
+            A result object which contains the query if it was successful,
+                or an error message.
+
+        Examples:
+            >>> kili.count_api_keys(user_id=user_id)
```
marcenacp, line 132: Sometimes examples are escaped in ```, sometimes not. Which option is best?

docs/index.md
```diff
@@ -22,18 +16,21 @@ Install the Kili client with pip:
...
+- Add the `KILI_API_KEY` variable in your bash environment (or in the settings of your favorite IDE) by pasting the API key value you copied above:
 
-```python
-from kili.client import Kili
-kili = Kili()
-# You can now use the Kili client!
-```
+  ```bash
+  export KILI_API_KEY='<you api key value here>'
+  ```
```
PierreLeveau, line 26: Why removing the doc about instantiating Kili without the API key ?

marcenacp (approved):

PierreLeveau (changes requested):
one small change (but great work!)

kili/queries/api_key/__init__.py
theodu: By default "Examples" section is displayed as a code block.
The cases where a wrap the example in ```is when I want to display a comment above the command line. Example:
```
Examples:
# List all issues of a project and their authors
 >>> kili.issues(project_id=project_id, fields=['author.email'])
```
will be displayed as below because he interpreted the # as a markdown h1
Capture d’écran 2022-05-03 à 10 33 24

The solution that I found was to force it into a code block by wrapping it inside a ```

docs/index.md
theodu: indeed, I'll change that

kili/queries/api_key/__init__.py
theodu: I just moves the comment on the same line as the command, this resolves the problem

PierreLeveau (approved):

PR #552, opened 2022-05-11
remove X-powered-By from graphQL queries

theodu:


theodu:
We saw with Aurélien, it is used nowhere. and the apollo client filter this header.
He suggested to rather push a 'User-Agent" header but this can be done later because I'll need to check that it doesn't break anything as it would be a new feature

PierreLeveau:
I agree, let's put that in User-Agent later. This will be useful for telemetry.

PierreLeveau (approved):

mDuval1 (approved):

PR #575, opened 2022-06-03
AAU, on the SDK, when I call kili.get_project, I instantiate a project object with high level methods

theodu:
…e client. Move exceptions in a specific file

### DIFF
--- a/kili/client.py
+++ b/kili/client.py
@@ -88,12 +88,12 @@ class Kili(  # pylint: disable=too-many-ancestors
         raise:
             NotFound if the given `project_id` does not correspond to an existing project
         """
-        query = self.projects(project_id=project_id,
-                              disable_tqdm=True, fields=['inputType', 'title'])
+        projects_response = self.projects(project_id=project_id,
+                                          disable_tqdm=True, fields=['inputType', 'title'])
 
-        if len(query) == 0:
+        if len(projects_response) == 0:
             raise NotFound(str(project_id))
-        project_fields = query[0]
+        project_fields = projects_response[0]
         title = project_fields['title']
         input_type = project_fields['inputType']
         return Project(client=self, project_id=project_id, input_type=input_type, title=title)
--- a/kili/client.py
+++ b/kili/client.py
@@ -3,6 +3,7 @@ This script permits to initialize the Kili Python SDK client.
 """
 import os
 
+from kili.exceptions import NotFound
 from kili.mutations.api_key import MutationsApiKey
 from kili.mutations.asset import MutationsAsset
 from kili.mutations.label import MutationsLabel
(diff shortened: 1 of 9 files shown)


test/test_client_project.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+
+    def test_get_unexisting_project(self):
+        """Test that get_project on a unexisting project_id raises a NotFound Error."""
+        with pytest.raises(NotFound):
+            kili.get_project('this_project_id_do_not_exist')
+
+    def test_get_existing_project(self, create_video_project):
+        """Test that get_project on an existing project_id return a Project with the right id."""
+        project_id = create_video_project
```
PierreLeveau, line 30: I'm against keeping digging the technical debt of testing on project that are hosted on the cloud. Can you mock the `kili.projects` method, as it is done in AutoML ?

PierreLeveau (changes requested):

test/test_client_project.py
theodu: Yes that's true, I'll change that

kili/client.py
```diff
@@ -75,3 +77,23 @@ def __init__(self, api_key=os.getenv('KILI_API_KEY'),
...
+        """Return a project object corresponding to the project_id given.
+        The returned project object inherit from many methods for project management
+
+        Args:
+            project_id: id of the project to return
+
+        raise:
+            NotFound if the given `project_id` does not correspond to an existing project
+        """
+        query = self.projects(project_id=project_id,
```
mDuval1, line 91: this is not a query but a response

kili/helpers.py
```diff
@@ -13,6 +13,8 @@
 
 import requests
 
+from kili.exceptions import EndpointCompatibilityError, GraphQLError
```
mDuval1, line 16: good to split in coherent files

kili/project/__init__.py
```diff
@@ -0,0 +1,16 @@
+"""Project module."""
+
+
+class Project:  # pylint: disable=too-few-public-methods
+    """
+    Object that represents a project in the database.
```
mDuval1, line 6: NOT `in the database` (not a concern of DS), but in `Kili`

PierreLeveau (approved):

kili/project/__init__.py
```diff
@@ -0,0 +1,16 @@
+"""Project module."""
+
+
+class Project:  # pylint: disable=too-few-public-methods
```
mDuval1, line 4: I don't think you should directly have a folder "project" at the root, but it depends on what you want to do. Do we intend to have a folder dataset, asset, label, etc... ? If so then we should have a folder above, named `entities` or something along those lines

theodu: That is what has been seen with Pierre and Edouard. It comes into the API v3 that we would like to deliver in the next quarter with a better structure, better UX and better function names. one idea would be to have several modules, one with project for example, to work on a specific project, one for automl later, one for managing its organization etc. This creates better scaling of the SDK and allows to work more easily with projects.
A more precise spec will come later, after the CLI
```
project = kili.get_project('<project_id>')
project.import(files)
```

PR #579, opened 2022-06-10
feat: implementation of append mutation for Issue and Comment

edarchimbaud:


PierreLeveau (changes requested):
spotted mainly doc issues

kili/mutations/comment/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+    def append_to_comments(
+        self,
+        text: str,
+        in_review: bool,
+        issue_id: str,
+    ):
+        """Create a comment.
+
+        Args:
+            text :
```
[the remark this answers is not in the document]
theodu: And I think that there should not be any space between the arg name and the colon otherwise the doc package doesn't understand it

kili/mutations/issue/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+        issue_number: int,
+        label_id: str,
+        object_mid: Optional[str],
+        type_: str,
+        external_id: str,
+        project_id: str,
+    ):
+        """Create a issue.
+
+        Args:
```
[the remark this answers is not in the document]
theodu: And I think that there should not be any space between the arg name and the colon otherwise the doc package doesn't understand it

kili/mutations/issue/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    def append_to_issues(
+        self,
+        issue_number: int,
+        label_id: str,
+        object_mid: Optional[str],
+        type_: str,
+        external_id: str,
+        project_id: str,
+    ):
+        """Create a issue.
```
theodu, line 36: create an issue

kili/mutations/issue/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+        """Initialize the subclass.
+
+        Args:
+            auth: KiliAuth object
+        """
+        self.auth = auth
+
+    @Compatible(["v1", "v2"])
+    @typechecked
+    def append_to_issues(
```
theodu, line 27: @PierreLeveau All mutations should be by batch to uniformise all methods, what do you think ?

PierreLeveau: agreed

theodu: Cannot append in batch because the resolver does not support it

theodu (approved):

kili/mutations/issue/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __init__(self, auth):
+        """Initialize the subclass.
+
+        Args:
+            auth: KiliAuth object
+        """
+        self.auth = auth
+
+    @Compatible(["v1", "v2"])
```
mDuval1, line 27: We should remove this decorator, it's one year old and not useful anymore

kili/mutations/issue/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    @typechecked
+    def append_to_issues(
+        self,
+        label_id: str,
+        project_id: str,
+        object_mid: Optional[str] = None,
+        text: Optional[str] = None,
+        type_: Literal["ISSUE", "QUESTION"] = "ISSUE",
```
mDuval1, line 35: is there a better name ? `is_issue` for example ?

theodu: The data quality team plan to add other types of issue (ex: warnings, ...) so a boolean does not seem not feasible

kili/mutations/issue/__init__.py
theodu: @mDuval1 Ok, what was this feature for ?
We can completely delete all the `Compatible` class and all its logic ?

kili/mutations/issue/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+            text: If given, write a comment related to the issue
+            project_id: Id of the project
+
+        Returns:
+            A result object which indicates if the mutation was successful,
+                or an error message.
+        """
+        issue_number = get_issue_number(self.auth, project_id, type_)
+        try:
+            asset_id = QueriesLabel(self.auth).labels(
```
PierreLeveau, line 52: cast to list before taking first element

kili/mutations/issue/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+        }
+        result = self.auth.client.execute(GQL_APPEND_TO_ISSUES, variables)
+        formated_result = format_result("data", result)
+        issue_id = formated_result["id"]
+
+        if text:
+            variables = {
+                "data": {
+                    "text": text,
+                    "inReview": False,
```
PierreLeveau, line 77: `inReview` should be a parameter to the method? If deprecated but mandatory to pass, you can add a comment to indicate that

kili/mutations/issue/helpers.py
```diff
@@ -0,0 +1,36 @@
...
+
+from ...queries.issue import QueriesIssue
+
+
+def get_issue_number(auth, project_id, type_):
+    """Get the next available issue number
+
+    Args:
+        auth: Kili Auth
+        project_id: Id of the project,
```
PierreLeveau, line 14: rm comma

kili/mutations/issue/helpers.py
```diff
@@ -0,0 +1,36 @@
...
+        type_: type of the issue to add
+    """
+    kili = QueriesIssue(auth)
+    issues = kili.issues(project_id=project_id, disable_tqdm=True, fields=["type", "issueNumber"])
+    if type_ == "ISSUE":
```
PierreLeveau, line 19:
```
issue_number = (
            max(
                [issue["issueNumber"] for issue in issues if issue["type"] == _type],
                default=-1,
            )
            + 1
        )
```

kili/mutations/issue/helpers.py
```diff
@@ -0,0 +1,36 @@
+"""
+Helpers for the issue mutations
+"""
+
+
+from ...queries.issue import QueriesIssue
+
+
+def get_issue_number(auth, project_id, type_):
```
PierreLeveau, line 9: add types

kili/mutations/issue/helpers.py
PierreLeveau: (to remove the `if type_ ... `)

PierreLeveau (changes requested):

kili/mutations/issue/__init__.py
theodu: This is quite tricky, inReview is a mandatory parameter but only useful when calling the append_to_comments mutation from the front. It does not have any legitimity on the SDK. 
In the front, it determines if the comment is posted in the review panel by a person with review rights. If yes, and if the comment has a question type, then it tags the comment as the answer of a question. 
Here, as the comment is the first comment to be posted after the issue/question creation, it is not an answer and there is no review process in the SDK so I always keep it to false. 
If one day we want to be able to add answers to comments though the SDK, we will have to think about that. ( Edouard first wrote a append_to_comments function in the SDK but I deleted it because Damien didn't need it and because this case makes the function ambigious in some cases)

PierreLeveau:
```suggestion
                    "inReview": False,  # mandatory and always False (can be True when used in the frontend
```

PierreLeveau (approved):

PR #582, opened 2022-06-14
AAU, on the CLI I see a function to list all projects with minimum information

FlorianLeRoyKili:

### DIFF
--- a/kili/cli.py
+++ b/kili/cli.py
@@ -26,7 +26,11 @@ def project():
 
 @project.command(name="list")
 @click.option('--api-key', type=str, envvar='KILI_API_KEY', required=True,
-              help='Your Api Key')
+              help=(
+                  'Your Kili API key (overrides the KILI_API_KEY environment variable).'
+                  'If not passed, requires the KILI_API_KEY environment variable to be set.'
+              )
+              )
 @click.option('--endpoint', type=str,
               default='https://cloud.kili-technology.com/api/label/v2/graphql',
               help='The API Endpoint')
@@ -65,7 +69,11 @@ def list_project(api_key: str,
 @project.command(name='import')
 @click.argument('files', type=click.Path(), nargs=-1, required=True)
 @click.option('--api-key', type=str, envvar='KILI_API_KEY', required=True,
-              help='Your Api Key')
+              help=(
+                  'Your Kili API key (overrides the KILI_API_KEY environment variable).'
+                  'If not passed, requires the KILI_API_KEY environment variable to be set.'
+              )
+              )
 @click.option('--endpoint', type=str,
               default='https://cloud.kili-technology.com/api/label/v2/graphql',
               help='The API Endpoint')
--- a/kili/cli.py
+++ b/kili/cli.py
@@ -50,8 +50,10 @@ def list_project(api_key: str,
     projects['progress'] = round(
         (1 - projects['numberOfRemainingAssets'] / projects['numberOfAssets']) * 100, 1)
     projects = projects[['title', 'id', 'progress', 'description']]
+    # Add '%' to progress if progress is not nan
     projects['progress'] = [(str(progress) + '%') if progress >=
                             0 else progress for progress in projects['progress']]
+    # If description or title has more than 50 characters, truncate after 47 and add '...'
     projects['description'] = [(description[:47] + '...') if len(description)
                                > 50 else description for description in projects['description']]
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -3,6 +3,7 @@ pandas
 pyparsing
 requests
 six
+tabulate
```
theodu, line 6: you also need to add it in the setup.py
requirements.txt is just an info for the users of the package that are required but they are not installed when doing a pip install kili. The packages in the setup.py are installed when installing Kili

test/test_cli.py
```diff
@@ -1,19 +1,40 @@
...
+                 'numberOfAssets': 10, 'numberOfRemainingAssets': 0}]
+
+
+def test_list(mocker):
+    mocker.patch("kili.client.Kili.__init__", return_value=None)
+    mocker.patch("kili.client.Kili.projects",
+                 side_effect=mocked__projects)
+
+    runner = CliRunner()
+    with runner.isolated_filesystem():
```
theodu, line 32: no need of this line, it creates a specific folder that will be deleted at the end of the test. here you write or write any files

test/test_cli.py
```diff
@@ -1,19 +1,40 @@
...
+        result = runner.invoke(list_project)
+        assert result.exit_code == 0
+        assert ((result.output.count("100.0%") == 1) and
+                (result.output.count("0.0%") == 2) and
+                (result.output.count("nan") == 1))
```
theodu, line 38: you can use the assert_called_once on the mocked project query. cf create_project test function

kili/cli.py
```diff
@@ -22,6 +24,39 @@ def project():
...
+              help='The API Endpoint')
+@click.option('--max', 'first', type=int, help='Maximum number of project to display', default=100)
+@click.option('--format', 'tablefmt', type=str, default='simple',
+              help='Defines how the table is formatted (see tabulate format).')
+def list_project(api_key: str,
+                 endpoint: str,
+                 tablefmt: str,
+                 first: int):
+    """
+    Print a list of projects in a table format
```
theodu, line 41: List your projects

theodu: and add an example

test/test_cli.py
FlorianLeRoyKili: I have tried a few ways but none worked, on which mock_patch am I suppose to call assert_called_once?

theodu: mock__projects = mocker.patch("kili.client.Kili.projects",
                 side_effect=mocked__projects)
...
mock__projects.assert_called_once()

I think that you have to be careful not to call it on `mocked__projects` which is just a function defined to replace the return of projects query. Just find a name for the mock, I called it mock__projects above

FlorianLeRoyKili: If you do this you have two mock__projects, the function and the mocker.patch. Is it OK ?

FlorianLeRoyKili: I have tried and apparently it is not. And it does not work either when I choose another name... Is it really necessary ? I don't see how I could call the function more than once.

theodu: Yes, I think that it is ok. 
The tests are also to keep things well in the long time and to avoid someone to decrease performances. if someone for example call one day the query a second time...
Things are good as they are, it was only a very small improvement

theodu (approved):

kili/cli.py
```diff
@@ -22,6 +24,42 @@ def project():
...
+    kili = Kili(api_key=api_key, api_endpoint=endpoint)
+    projects = kili.projects(fields=['title', 'id', 'description', 'numberOfAssets',
+                             'numberOfRemainingAssets', 'numberOfReviewedAssets'], first=first)
+    projects = pd.DataFrame(projects)
+    projects['progress'] = round(
+        (1 - projects['numberOfRemainingAssets'] / projects['numberOfAssets']) * 100, 1)
+    projects = projects[['title', 'id', 'progress', 'description']]
+    projects['progress'] = [(str(progress) + '%') if progress >=
+                            0 else progress for progress in projects['progress']]
+    projects['description'] = [(description[:47] + '...') if len(description)
```
PierreLeveau, line 55: can you add in comment where does this 47 comes from?

PierreLeveau (approved):

PierreLeveau:
Note: no branch has been created on the `kili` Gitlab side as the interactions with the backend have not been touched.

PR #590, opened 2022-06-27
AAU, on the SDK, when I call kili.add_assets_to_review(...), I want to add these assets to review

theodu:

### DIFF
--- a/kili/mutations/asset/__init__.py
+++ b/kili/mutations/asset/__init__.py
@@ -8,7 +8,9 @@ from typeguard import typechecked
 
 from ...helpers import (Compatible, format_result)
 from ...queries.project import QueriesProject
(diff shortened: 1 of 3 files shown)


kili/mutations/asset/__init__.py
```diff
@@ -249,3 +252,38 @@ def generate_variables(batch):
                                               generate_variables,
                                               GQL_DELETE_MANY_FROM_DATASET)
         return format_result('data', results[0], Asset)
+
+    @Compatible(['v1', 'v2'])
+    @typechecked
+    def add_assets_to_review(
```
p-desaintchamas, line 258: Why not use the dedicated resolver in backend `addAllLabeledAssetsToReview` that ensures the assets have a label of type default and their status is Labeled ?

theodu: @p-desaintchamas I saw that resolver but with its name I thought that it was appending all labeled assets to review. this is not the case ?

p-desaintchamas (approved):

PR #594, opened 2022-06-28
AAU, on the CLI, when I create a project, I see that I can import the json_interface from another project

FlorianLeRoyKili:

### DIFF
--- a/kili/cli.py
+++ b/kili/cli.py
@@ -31,7 +31,6 @@ endpoint_option = click.option(
     help='The API Endpoint.'
 )
 
-
 tablefmt_option = click.option('--stdout-format', 'tablefmt', type=str, default='plain',
                                help='Defines how the output table is formatted '
                                '(see https://pypi.org/project/tabulate/, default: plain).')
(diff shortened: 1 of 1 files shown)


PierreLeveau (approved):

kili/cli.py
```diff
@@ -83,8 +89,12 @@ def list_project(api_key: str,
 @project.command(name='create')
 @api_key_option
 @endpoint_option
-@click.option('--interface', type=click.Path(exists=True), required=True,
-              help="Path pointing to your json interface file.")
+@click.option('--interface', type=str, required=True,
+              help=(
+                  "Path pointing to your json interface file."
+                  "of project_id of another Kili project"
```
theodu, line 95: Path pointing to your json interface file or the project_id of another Kili project

kili/cli.py
```diff
@@ -108,13 +118,27 @@ def create_project(api_key: str,
             --input-type TEXT \\
             --title "Invoice annotation project"
         ```
-
+        ```
```
theodu, line 121: we should add a comment to say that it will create a new project with the same json_interface.
It can be ambiguous wether it also add the assets and the members to...

theodu (changes requested):
little typo and missing doc update

theodu (approved):

PR #601, opened 2022-06-30
feature/ml 526 aamle on pythonsdk when i do not pass project id deprecation warning

PierreLeveau:
- Revert "chore(ML-268): make project_id mandatory in queries"
- feat: deprecate project_id = None

### DIFF
--- a/kili/queries/asset/__init__.py
+++ b/kili/queries/asset/__init__.py
@@ -32,7 +32,7 @@ class QueriesAsset:
     # pylint: disable=dangerous-default-value
     @Compatible(['v1', 'v2'])
     @typechecked
-    @deprecate(removed_in="2.115")
+    @deprecate(removed_in="2.117")
     def assets(self,
                asset_id: Optional[str] = None,
                project_id: Optional[str] = None,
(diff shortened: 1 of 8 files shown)


kili/queries/asset/__init__.py
```diff
@@ -236,9 +236,8 @@ def _query_assets(self,
     @Compatible(['v1', 'v2'])
     @typechecked
     @deprecate(removed_in="2.115")
```
theodu, line 238:
```suggestion
    @deprecate(removed_in="2.117")
```

theodu (changes requested):

kili/queries/asset/__init__.py
theodu: This has to be changed everywhere. otherwise the bump to 2.115 will be blocked.
We want to block it in the 2.117 if we forgot to remove the deprecation warning

kili/queries/asset/__init__.py
```diff
@@ -34,8 +34,8 @@ def __init__(self, auth):
...
     def assets(self,
-               project_id: str,
                asset_id: Optional[str] = None,
+               project_id: Optional[str] = None,
```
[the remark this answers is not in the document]
theodu: they are not launched in all the mutations are queries

kili/queries/asset/__init__.py
PierreLeveau: done

theodu (approved):

PR #602, opened 2022-07-04
fix: deprecation message error

PierreLeveau:

### DIFF
--- a/kili/queries/asset/__init__.py
+++ b/kili/queries/asset/__init__.py
@@ -172,7 +172,7 @@ class QueriesAsset:
 
         saved_args = locals()
         count_args = {k: v for (k, v) in saved_args.items()
-                      if k not in ['skip', 'first', 'disable_tqdm', 'format', 'fields', 'self', 'as_generator']}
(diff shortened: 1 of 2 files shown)


kili/queries/asset/__init__.py
```diff
@@ -172,7 +172,7 @@ def assets(self,
 
         saved_args = locals()
         count_args = {k: v for (k, v) in saved_args.items()
-                      if k not in ['skip', 'first', 'disable_tqdm', 'format', 'fields', 'self', 'as_generator']}
```
theodu, line 175: to avoid such errors, shouldn't we remove the `count_args = {k: v for (k, v) in saved_args.items()`and rather write the count_args explicitely ? And same in other queries and mutations ?

FlorianLeRoyKili (approved):

kili/queries/asset/__init__.py
PierreLeveau: yes we should, maybe not right now. We should refactor the all pagination + count handling + formatting + ... in a single class. We are not DRY at all

theodu: agree

PR #604, opened 2022-07-04
AAU, on the SDK, when I call kili.send_back_to_queue(..., I want to send assets back to queue

FlorianLeRoyKili:

### DIFF
--- a/kili/mutations/asset/__init__.py
+++ b/kili/mutations/asset/__init__.py
@@ -258,7 +258,7 @@ class MutationsAsset:
 
     @Compatible(['v1', 'v2'])
     @typechecked
-    def add_assets_to_review(
+    def add_to_review(
             self,
             asset_ids: List[str]) -> dict:
         """Add assets to review.
@@ -274,7 +274,7 @@ class MutationsAsset:
                 or an error message.
 
         Examples:
-            >>> kili.add_assets_to_review(
+            >>> kili.add_to_review(
                     asset_ids=[
                         "ckg22d81r0jrg0885unmuswj8",
                         "ckg22d81s0jrh0885pdxfd03n"
@@ -293,13 +293,10 @@ class MutationsAsset:
 
     @Compatible(['v2'])
     @typechecked
-    def send_back_assets_to_queue(
+    def send_back_to_queue(
             self,
             asset_ids: List[str]):
         """Send assets back to queue.
-
-        !!! warning
-            Assets without status REVIEWED will be ignored
 
         Args:
             asset_ids: The asset IDs to add to review
--- a/kili/mutations/asset/__init__.py
+++ b/kili/mutations/asset/__init__.py
@@ -11,11 +11,12 @@ from ...queries.project import QueriesProject
 from .queries import (GQL_ADD_ALL_LABELED_ASSETS_TO_REVIEW,
                       GQL_APPEND_MANY_FRAMES_TO_DATASET,
                       GQL_DELETE_MANY_FROM_DATASET,
-                      GQL_UPDATE_PROPERTIES_IN_ASSETS)
+                      GQL_UPDATE_PROPERTIES_IN_ASSETS,
+                      GQL_SEND_BACK_ASSETS_TO_QUEUE)
 from .helpers import (process_append_many_to_dataset_parameters,
                       process_update_properties_in_assets_parameters)
 from ...constants import NO_ACCESS_RIGHT
-from ...orm import Asset, AssetStatus
+from ...orm import Asset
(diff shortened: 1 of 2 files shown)


kili/mutations/asset/__init__.py
```diff
@@ -289,3 +290,38 @@ def generate_variables(batch):
...
+        properties_to_batch = {'asset_ids': asset_ids}
+
+        def generate_variables(batch):
+            return {'where': {'idIn': batch['asset_ids']}}
+
+        results = _mutate_from_paginated_call(self,
+                                              properties_to_batch,
+                                              generate_variables,
+                                              GQL_SEND_BACK_ASSETS_TO_QUEUE)
+        return format_result('data', results[0])
```
PierreLeveau, line 327: is there something preventing us to return all the results?

PierreLeveau: it may be done that way in other methods, but we lose some information doing that

theodu: all mutations return an object with the project ID.
To keep it coherent with before, we still return an object with the project ID if all mutations succeed.
If all mutations succeed, results have all the same objects so we just return the first one.

theodu: But it would be interesting to think again about the mutations results...
Like giving the ids of assets changed and not the project id for example

theodu (changes requested):

kili/mutations/asset/__init__.py
FlorianLeRoyKili: exactly. Basically the full results variable is something like : data : {data : {id : project_id}}, so not much to return except the project id

kili/mutations/asset/__init__.py
```diff
@@ -289,3 +290,38 @@ def generate_variables(batch):
                                               generate_variables,
                                               GQL_ADD_ALL_LABELED_ASSETS_TO_REVIEW)
         return format_result('data', results[0])
+
+    @Compatible(['v2'])
+    @typechecked
+    def send_back_assets_to_queue(
```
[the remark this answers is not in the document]
FlorianLeRoyKili: I agree the name "send_assets_back_to_queue" seems best to me but the resolver is "sendBackAssetsToQueue", that's why I used "send_back_assets_to_queue". Wdyt ?

kili/mutations/asset/__init__.py
```diff
@@ -289,3 +290,38 @@ def generate_variables(batch):
...
+            self,
+            asset_ids: List[str]):
+        """Send assets back to queue.
+
+        !!! warning
+            Assets without status REVIEWED will be ignored
```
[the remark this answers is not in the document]
FlorianLeRoyKili: You're right I forgot to check. Stand_by...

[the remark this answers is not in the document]
FlorianLeRoyKili: Ok, so unfortunately it is not true. Assets are sent to status "ONGOING" even if they did not have the status "REVIEWED" before (all assets with labels are sent to status ONGOING). I suggest another warning: 
`!!! warning'
'            Assets will be sent back to queue even if they have not been reviewed`

kili/mutations/asset/__init__.py
PierreLeveau: yes this should be done for v3 imho (probably a list to stay consistent with the input)

kili/mutations/asset/__init__.py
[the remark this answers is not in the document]
theodu: @p-desaintchamas last time that I was in a call with Damien to speak about this, we agreed that we are sending a label back to queue and not an asset. I see that the resolver and the app rather talk about assets sent back to queue. How did you manage this for the final users ?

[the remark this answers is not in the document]
FlorianLeRoyKili: I don't see how we could send a label back to queue? I mean, here, we changed the asset_status from "REVIEWED" to "ONGOING" (label_type remains "REVIEW"). For a label, it would mean changing the label_type from "REVIEW" to "DEFAULT" I guess, but what would you do about the asset_status?

kili/mutations/asset/__init__.py
```diff
@@ -257,7 +258,7 @@ def generate_variables(batch):
 
     @Compatible(['v1', 'v2'])
     @typechecked
-    def add_assets_to_review(
+    def add_to_review(
```
theodu, line 261: also modify the name of the function in the Examples in the docstring

theodu (approved):

PierreLeveau (approved):

PR #607, opened 2022-07-05
feat: clean the exception strings

PierreLeveau:
first step to clean up errors in the SDK: get rid of the stack trace coming from the backend, and only keep the message.

### DIFF
--- a/kili/exceptions.py
+++ b/kili/exceptions.py
@@ -1,5 +1,5 @@
 """Exceptions of the package."""
-
+import ast
 
 class NotFound(Exception):
     """Used when a given object is not found in Kili"""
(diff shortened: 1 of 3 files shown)


theodu:
Already improves a lot!!

theodu:
Capture d’écran 2022-07-06 à 10 43 32
Capture d’écran 2022-07-06 à 10 41 48
Could be also get rid of all the traceback before ? It might be useful for a developer for clients this is useless. Isn't it possible with only `sys.tracebacklimit=0` ?

theodu (commented):
Could be also get rid of all the traceback before ? It might be useful for a developer for clients this is useless. Isn't it possible with only `sys.tracebacklimit=0` ?

PierreLeveau:
> Could be also get rid of all the traceback before ? It might be useful for a developer for clients this is useless. Isn't it possible with only `sys.tracebacklimit=0` ?

i'm less shoked by the client-side errors (maybe my Python bias?), it provides some context and is not unusual (having server-side stacktrace is unusual)

PierreLeveau:
asking for @crsegerie advice, wdyt?

theodu:
Yes, I see, but for me, traceback is useful for someone that tries to debug its script if he has errors. A client won't debug the SDK, it is our role. It would be useful for us while developing but send a hard-to-read error for the client, don't you think ?

PierreLeveau:
For CLI, I can agree with you, for the SDK, it is usual to keep the stack trace (especially for open-source code)

PR #608, opened 2022-07-06
add callback to accept list of files in --exclude option to project import 

FlorianLeRoyKili:


PierreLeveau:
files is probably not a good example to use csv strings instead of multiple, as a filename can contain a comma. wdyt?

FlorianLeRoyKili:
I vote we kill it. things are OK as they are I think.

PierreLeveau:
or they could also exclude through a regexp

PierreLeveau:
canceled

PR #610, opened 2022-07-06
AAU, on the CLI, I see a better structure in the code

FlorianLeRoyKili:

### DIFF
--- a/kili/cli/common_args.py
+++ b/kili/cli/common_args.py
@@ -1,10 +1,12 @@
+"""Common arguments and options for the CLI"""
 import click
 
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
 
-class Options:
+class Options:  # pylint: disable=too-few-public-methods
+    """Common options for the CLI"""
 
     api_key = click.option(
         '--api-key', type=str, default=None,
--- a/kili/cli/project/__init__.py
+++ b/kili/cli/project/__init__.py
@@ -1,14 +1,14 @@
+"""Project command of Kili CLI"""
+
 import click
 
 from kili.cli.common_args import CONTEXT_SETTINGS
 
-from kili.cli import project
-
-from kili.cli.project.create import create
-from kili.cli.project.describe import describe
-from kili.cli.project.import_ import import_
-from kili.cli.project.label import label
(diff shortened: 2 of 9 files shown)


kili/cli/cli.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@cli.group(context_settings=CONTEXT_SETTINGS)
+def project():
+    """Commands to interact with a Kili project"""
+
+
+project.add_command(create, name="create")
+project.add_command(describe, name="describe")
+project.add_command(import_assets, name="import")
```
PierreLeveau, line 31: why not renaming the python functions into the command names? this would be even clearer.

PierreLeveau (changes requested):
2 suggestions

kili/cli/cli.py
FlorianLeRoyKili: You mean : `from kili.cli.project.create import create_project` then `project.add_command(create_project, name="create")`. If so, then I agree.

kili/cli/cli.py
```diff
@@ -0,0 +1,47 @@
...
+@click.version_option(__version__)
+def cli():
+    """Kili Command line Interface
+
+       To get all the available commands, please type: `kili project --help`.
+    """
+
+
+@cli.group(context_settings=CONTEXT_SETTINGS)
+def project():
```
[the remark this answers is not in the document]
FlorianLeRoyKili: Apparently use add_command also works with group so we're OK !

kili/cli/cli.py
PierreLeveau:
```
from kili.cli import project
project.add_command(project.create, name="create")
```

theodu:
@FlorianLeRoyKili you miss the __init__ in the CLI folder. I already had this problem where I forget to add a __init__ in a new folder, it was working in local but not when packaged.

theodu (changes requested):
@FlorianLeRoyKili you miss the init in the CLI folder. I already had this problem where I forget to add an init in a new folder, it was working in local but not when packaged.

theodu (changes requested):

PierreLeveau:
Linting issue, some you can ignore I think

theodu (approved):

PR #613, opened 2022-07-08
chore: separate notebook tests and run on weekdays

PierreLeveau:


PierreLeveau:
the scheduled one does not seem to have, will investigate

PierreLeveau:
it seems to apply on the master branch only, so I'll merge first

PR #619, opened 2022-07-12
AAU, on model Label, I see field skipped is deprecated

gheorghetutunaru:

### DIFF
--- a/kili/queries/label/__init__.py
+++ b/kili/queries/label/__init__.py
@@ -73,6 +73,8 @@ class QueriesLabel:
             created_at_lte: Returned labels should have a label whose creation date is lower than this date.
             fields: All the fields to request among the possible fields for the labels.
                 See [the documentation](https://docs.kili-technology.com/reference/graphql-api#label) for all possible fields.
+                Notice that the field "skipped" is deprecated since: 20/07/2022.
+                It will be removed after: 31/08/2022.
             first: Maximum number of labels to return.
             honeypot_mark_gte: Returned labels should have a label whose honeypot is greater than this number.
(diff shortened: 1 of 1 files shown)


kili/queries/label/__init__.py
```diff
@@ -44,7 +44,7 @@ def labels(self,
                created_at_gte: Optional[str] = None,
                created_at_lte: Optional[str] = None,
                fields: List[str] = ['author.email', 'author.id', 'id',
-                                    'jsonResponse', 'labelType', 'secondsToLabel', 'skipped'],
```
theodu, line 47: removing it from default fields can be breaking. The purpose to show deprecation warning is not warn clients before to make it breaking. We should wait until 31/08/2022 to remove it here.

kili/queries/label/__init__.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, auth):
     # pylint: disable=dangerous-default-value
     @Compatible(['v1', 'v2'])
     @typechecked
-    @deprecate(removed_in="2.116")
+    @deprecate(removed_in="2.118")
```
theodu, line 37: This is tool to remember to remove the deprecation warning in the 2.116 (which is here the project_id), we should stick to `@deprecate(removed_in="2.116")` and change it to 2.118 once that we will have removed the first deprecation

theodu (changes requested):

kili/queries/label/__init__.py
gheorghetutunaru: Oh, okay, thank you. Added it back

PR #632, opened 2022-07-26
AAKD, on Python SDK, when I commit, It runs isort

FlorianLeRoyKili:


setup.cfg
```diff
@@ -1,3 +1,9 @@
 # Inside of setup.cfg
 [metadata]
 description-file = README.md
+
+[isort]
+# https://pycqa.github.io/isort/docs/configuration/options.html
+line_length = 120
```
PierreLeveau, line 7:
```suggestion
line_length = 100
```
we have set 100 in the black config

PierreLeveau (changes requested):

PierreLeveau (approved):

PR #645, opened 2022-08-24
AA KD, on Apollo, I see a dashboard with the performance latency by clients/fields

theodu:


PierreLeveau (changes requested):
we should do the full story and add the possibility to pass the client name to the `GraphQLClient` (`Literal["kili-python", "kili-cli"]`) through the `Kili` client instantiation (default being `kili-python`)

kili/graphql_client.py
```diff
@@ -12,14 +12,30 @@
 import websocket
 from six.moves import urllib
 
+from . import __version__
+
+
+class GraphQLClientName:  # pylint: disable=too-few-public-methods
```
PierreLeveau, line 18:
```
from enum import Enum

class GraphQLClientName(Enum):
```

kili/cli/project/create.py
```diff
@@ -63,7 +64,7 @@ def create_project(
     To build a Kili project interface, please visit: \n
     https://docs.kili-technology.com/docs/customizing-the-interface-through-json-settings
     """
-    kili = Kili(api_key=api_key, api_endpoint=endpoint)
+    kili = Kili(api_key=api_key, api_endpoint=endpoint, client_name=GraphQLClientName.CLI)
```
PierreLeveau, line 67: for this one, I think the best option is to create a helper cli function ` get_kili_client` that instantiates the kili client with the correct client info:
```
def get_kili_client(api_key: str, api_endpoint: str):
      return Kili(..., client_name=GraphQLClientName.CLI)
```
this hides a detail, and is more future proof if we need to do more adaptations to the client

PierreLeveau (changes requested):

PierreLeveau:
did you check that the provided information is correctly displayed in Apollo studio?

kili/graphql_client.py
```diff
@@ -12,14 +12,30 @@
...
+    """GraphQL client name."""
+
+    SDK = "python-sdk"
```
PierreLeveau, line 21: `kili-python` (the sdk is the client + recipes + some doc).

kili/graphql_client.py
```diff
@@ -12,14 +12,30 @@
...
+    CLI = "python-cli"
```
PierreLeveau, line 22: the fact the cli is coded in Python is an implementation detail, so I would say `kili-cli`

PierreLeveau (changes requested):

kili/graphql_client.py
utix: As there is (was) a node client, I think we can keep python

PierreLeveau (approved):

PR #648, opened 2022-08-30
feat(ML-630): allow to reduce number of trials

mDuval1:


kili/graphql_client.py
```diff
@@ -83,7 +84,8 @@ def _send(self, query, variables):
 
         if self.session is not None:
             try:
-                number_of_trials = 10
+                number_of_trials_env = os.getenv("KILI_SDK_TRIALS_NUMBER")
+                number_of_trials = number_of_trials_env if number_of_trials_env is not None else 10
```
PierreLeveau, line 88:
```suggestion
                number_of_trials = os.getenv("KILI_SDK_TRIALS_NUMBER", 10)
```

PierreLeveau (changes requested):

kili/graphql_client.py
mDuval1: I need to manage if the env variable value is not a string, I pushed new changes

PierreLeveau (approved):

antonioloison (commented):
good for me

PR #653, opened 2022-09-01
fix(COM-214): early return right away instead of waiting 1s

mDuval1:


hugo-mailfait-kili (approved):

hugo-mailfait-kili (commented):
I think you also need to update the `setup.cfg` file if you want to bump the version and release

PierreLeveau (approved):
LGTM, but I have a more general question:
We assume that:
* 200 without errors = correct response
* 200 with errors = error
and that it is not possible to have something else than 200 and 401 ?

PR #655, opened 2022-09-02
Yolo export in CLI and SDK

PierreLeveau:
This PR adds the export feature to the CLI and the SDK. The Yolo formats have been implemented for now.
This is more or less a copy-paste of the code of `label-export`, with the following modifications:
* removal of the web-specific code
* change of the enum types to literal (more convenient for the end-user)
* `YoloExporterSelector` handles the exporter selection (split or merged layout) and initializes the dependencies. This is not a proper dependency injection, this will be done later.
* `YoloSplitExporter` and `YoloMergedExporter` handle the export operations and common code is factorized in the parent class.
* `ContentRepository` objects handle fetching the assets from the backend.
* A service layer has been added to handle the calls from the Python client through `project` and the CLI.

### DIFF
--- a/kili/services/export/typing.py
+++ b/kili/services/export/typing.py
@@ -1,0 +1,27 @@
+"""
+Types used by the conversion service
+"""
+from typing import NamedTuple, NewType, Tuple, Union
+
+from typing_extensions import Literal
+
+LabelFormat = Literal["yolo_v4", "yolo_v5"]
+InputType = Literal["TEXT", "IMAGE"]
+ExportType = Literal["latest", "normal"]
+AssetId = NewType("AssetId", str)
+ProjectId = NewType("ProjectId", str)
+SplitOption = Literal["split", "merged"]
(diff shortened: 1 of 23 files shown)


mDuval1 (approved):

kili/services/export/typing.py
```diff
@@ -0,0 +1,27 @@
+"""
+Types used by the conversion service
+"""
+from typing import NamedTuple, NewType, Tuple, Union
+
+from typing_extensions import Literal
+
+LabelFormat = Literal["yolo_v4", "yolo_v5"]
+InputType = Literal["text", "image"]
```
theodu, line 9: We should use this one everywhere. shouldn't we create a typing.py at the root of kili with common types like this one ?
And as the backend interpret it only in upper case, maybe have InputType = Literal["TEXT", "IMAGE", "PDF" , ect...]

theodu (changes requested):

theodu (approved):

PR #656, opened 2022-09-08
AAU, on the SDK, when I import assets, I see it is fast

theodu:


theodu:
Two imports are not done with signed URL yet: import video from local frames and import time series from CSV

kili/mutations/asset/helpers.py
```diff
@@ -104,18 +104,20 @@ def upload_content(content_array: List[str], input_type, auth, project_id):
     content_type_array = []
     for content in content_array:
         if os.path.exists(content) and check_file_mime_type(content, input_type):
-            data_array.append(open(content, "rb"))
+            with open(content, "rb") as binary_data:
+                data_array.append(binary_data)
```
PierreLeveau, line 108: the `binary_data` file pointer is closed, so you wont be able to handle it. If you want to read it later without creating the file pointer, you can simply store the filenames and open them when you want to read them.

kili/helpers.py
```diff
@@ -427,9 +427,9 @@ def check_file_mime_type(path: str, input_type: str, verbose: bool = True) -> bo
         return False
 
     correct_mime_type = mime_type in mime_extensions_for_IV2[input_type]
-    if verbose and not correct_mime_type:
-        print(
-            f"File mime type for {path} is {mime_type} and does not correspond"
+    if not correct_mime_type:
+        raise ValueError(
```
PierreLeveau, line 431: 👍

kili/mutations/asset/helpers.py
```diff
@@ -62,45 +68,74 @@ def get_file_mimetype(
     return None
 
 
-def process_json_content(
+def process_and_store_json_content(
     input_type: str,
     content_array: List[str],
     json_content_array: Union[List[str], None],
+    project_id,
```
PierreLeveau, line 75: add types to new arguments

kili/mutations/asset/helpers.py
```diff
@@ -62,45 +68,74 @@ def get_file_mimetype(
...
+        json_content_array = list(map(process_frame_json_content, json_content_array))
+    signed_urls = request_signed_urls(auth, project_id, len(json_content_array))
+    json_content_array = list(map(dumps, json_content_array))
+    return upload_data_via_rest(
+        signed_urls, json_content_array, ["text/plain"] * len(json_content_array)
+    )
 
 
-def process_content(
+def upload_content(content_array: List[str], input_type, auth, project_id):
```
PierreLeveau, line 95: Add types to arguments

kili/mutations/asset/helpers.py
```diff
@@ -314,3 +354,56 @@ def generate_json_metadata_array(as_frames, fps, nb_files, input_type):
...
+    """
+    payload = {
+        "projectID": project_id,
+        "size": size,
+    }
+    urls_response = auth.client.execute(GQL_CREATE_UPLOAD_BUCKET_SIGNED_URLS, payload)
+    return urls_response["data"]["urls"]
+
+
+def upload_data_via_rest(signed_urls, data_array: List[str], content_type_array: List[str]):
```
PierreLeveau, line 375: add type to signed_urls

PierreLeveau: note: I think that, moving forward, we should handle list of typed dicts instead of several arrays, until we call the resolvers where data can be split. I know this will require some back and forth to maintain the V2 API, but the code will be more readable and debuggable.

PierreLeveau (changes requested):
nice to have: add one service level test. We can discuss that once the PR is finished.

kili/mutations/asset/helpers.py
```diff
@@ -314,3 +348,58 @@ def generate_json_metadata_array(as_frames, fps, nb_files, input_type):
...
+    """Determine if all data to append is hosted data or local data.
+    If it is a mix of both, return an error
+    """
+    if json_content_array:
+        return False
+    if all(is_url(content) for content in content_array):
+        return False
+    if all(not is_url(content) for content in content_array):
+        return True
+    raise ValueError("Content to append should either be all hosted file or all local files")
```
PierreLeveau, line 405: you can specify an Error like `NonUniformContentHostingError`.

test/mutations/test_asset.py
```diff
@@ -105,11 +113,18 @@ def setUp(self):
         # Create a temporary directory
         self.test_dir = tempfile.mkdtemp()
 
-    def tearDown(self):
+    def tearDown(self, mocker):
         # Remove the directory after the test
         shutil.rmtree(self.test_dir)
 
-    def test_geotiff_upload_properties(self):
+        mocker.patch(
```
PierreLeveau, line 120: why did the method `test_geotiff_upload_properties` go away? `tearDown` is used to clean-up the test environment.

PierreLeveau (changes requested):

test/mutations/test_asset.py
theodu: that's a mistake, I put it back

kili/mutations/asset/helpers.py
theodu: Indeed! But with the new MR of maxime, this error and this function will be deleted. I take this in the next ticket that I take: https://linear.app/kili-technology/issue/ML-641/aau-on-the-sdk-when-i-use-append-many-to-dataset-i-see-files-are

PR #658, opened 2022-09-13
AAKD, on the SDK, when I call append_many_to_dataset, It calls an import_assets services

theodu:


kili/utils/typing.py
```diff
@@ -0,0 +1,25 @@
+"""Common typings """
+from enum import Enum
```
PierreLeveau, line 2: should be moved to `kili.typing` as Literals but Raphael is already moving things there,  we'll get in sync about that.

PierreLeveau (changes requested):
very good job! I'm curious about the results of :
```
PYTHONPATH=$(pwd) pytest --cov=kili/services/import_asset test/services/import_asset
```

PierreLeveau:
and it would make sense to rename the service to `import_assets`

PR #659, opened 2022-09-14
AAKD, on the SDK, when I call append_many_to_dataset, It calls an import_assets services and it uses signed urls

theodu:


kili/mutations/asset/__init__.py
```diff
@@ -98,43 +94,24 @@ def append_many_to_dataset(
...
+                "is_honeypot": is_honeypot_array and is_honeypot_array[i],
+            }
+            for i in range(nb_data)
+        ]
 
-        results = _mutate_from_paginated_call(
-            self, properties_to_batch, generate_variables, request
-        )
-        return format_result("data", results[0], Asset)
+        import_assets(self.auth, project_id=project_id, assets=assets)
```
PierreLeveau, line 114: you can replace `self.auth` by self, so you don't need to instanciate kili again.

kili/mutations/asset/queries.py
```diff
@@ -2,39 +2,9 @@
 Queries of asset mutations
 """
 
+from kili.graphql.operations.asset.fragments import ASSET_FRAGMENT
```
PierreLeveau, line 5: should be moved to operations/asset/fragments.py ?

kili/services/asset_import/legacy.py
```diff
@@ -0,0 +1,375 @@
...
+        self,
+        auth: KiliAuth,
+        project_id: str,
+        input_type: str,
+        assets: List[AssetToImport],
+    ):
+        self.input_type = input_type
+        self.project_id = project_id
+        self.auth = auth
+        self.content_array = assets[0].get("content", None) and [
```
PierreLeveau, line 43: it is a better practice to remove the arrays from the attributes, and make `assets` an argument of the `.import_assets` method: `def import(assets: List[AssetToImport])`

kili/services/asset_import/legacy.py
```diff
@@ -0,0 +1,375 @@
...
+        if is_url(json_content):
+            processed_json_content_array.append(json_content)
+        else:
+            if input_type in ("FRAME", "VIDEO"):
+                json_content = process_frame_json_content(json_content)
+            processed_json_content_array.append(dumps(json_content))
+    return processed_json_content_array
+
+
+def uplaod_content(signed_url: str, content: str, input_type: str):
```
PierreLeveau, line 164: upload_content

test/cli/test_project.py
```diff
@@ -157,7 +157,7 @@ def mocked__project_assets(project_id=None, **_):
 kili_client.append_to_labels = append_to_labels_mock = MagicMock()
 kili_client.create_predictions = create_predictions_mock = MagicMock()
 kili_client.count_projects = count_projects_mock = MagicMock(return_value=1)
-kili_client.append_many_to_dataset = append_many_to_dataset_mock = MagicMock()
+# kili_client.append_many_to_dataset = append_many_to_dataset_mock = MagicMock()
```
PierreLeveau, line 160: rm ?

PierreLeveau (changes requested):

kili/mutations/asset/__init__.py
theodu: but self is not `Kili` but `MutationAsset` here. So i won't be able to run other queries to get informations like the project `input_type` in `import_assets` because `projects` is a method of `Kili` and not `MutationAsset`
And in the new architecture, services won't call kili.projects but `get_input_type` (to code) that will not instantiate a new kili client, isn't it ?

kili/mutations/asset/queries.py
theodu: you mean the PROJECT_FRAGMENT_ID ?

kili/services/asset_import/legacy.py
theodu: good catch!

kili/mutations/asset/__init__.py
PierreLeveau: ok I did not see that. I think that each Mixin should have access to a client, since they don't have a reason to exist otherwise. To be discussed later then.

PierreLeveau:
you can rebase master before merging

PierreLeveau (approved):

PR #671, opened 2022-09-20
feat: fix append_many_to_dataset result return

theodu:


kili/mutations/asset/__init__.py
```diff
@@ -111,7 +111,8 @@ def append_many_to_dataset(
             for i in range(nb_data)
         ]
 
-        import_assets(self.auth, project_id=project_id, assets=assets)
+        result = import_assets(self.auth, project_id=project_id, assets=assets)
```
PierreLeveau, line 114: no need to create a new variable `return import_assets(self.auth, project_id=project_id, assets=assets)`

kili/services/asset_import/__init__.py
```diff
@@ -23,4 +23,5 @@ def import_assets(
     input_type = projects[0]["inputType"]
 
     legacy_importer = LegacyImporter(auth=auth, project_id=project_id, input_type=input_type)
-    legacy_importer.import_assets(assets=assets)
+    result = legacy_importer.import_assets(assets=assets)
```
PierreLeveau, line 26: same here

PierreLeveau (approved):
LGTM with a suggestion

kili/mutations/asset/__init__.py
```diff
@@ -111,7 +111,8 @@ def append_many_to_dataset(
...
+        return result
```
PierreLeveau, line 115:
```suggestion
        return import_assets(self.auth, project_id=project_id, assets=assets)
```

kili/services/asset_import/__init__.py
```diff
@@ -23,4 +23,5 @@ def import_assets(
...
+    return result
```
PierreLeveau, line 27:
```suggestion
    return legacy_importer.import_assets(assets=assets)
```

PR #672, opened 2022-09-21
AAKD, on Python SDK, when I run unit tests, they do not depend on a running server

theodu:


test/queries/test_asset.py
```diff
@@ -1,15 +1,15 @@
...
 
 import pandas as pd
 
-from kili.client import Kili
+from kili.queries.asset import QueriesAsset
 
-api_endpoint = os.getenv("KILI_API_ENDPOINT")
-kili = Kili(api_endpoint=api_endpoint)
+mocked_auth = MagicMock()
+mocked_auth.client.endpoint = "https://staging.cloud.kili-technology.com/api/label/v2/graphql"
```
PierreLeveau, line 12: we should put a random string here right? otherwise it will call the staging server

PierreLeveau (changes requested):

PR #684, opened 2022-09-27
feat: launch daily test on bump and add possibility to bump from a sp…

theodu:
- recipes workflow are now launched when on branches that begins by "release/" (same as in the script in entrypoint.sh)
- it is now possible to bump from a specific commit. If no commit is given, HEAD will be used by default
- the bump commit is now tagged with the release number

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -18,7 +18,7 @@ if [[ "$1" == "bump:"* ]]
 
     #create or reset release branch
     git checkout master
-    git pull origin master --tags
+    git pull origin master
     git checkout -B release/bump-version-$(date +%s) $COMMIT
 
     echo "Creating a new branch from master commit $COMMIT"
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -24,7 +33,8 @@ if [[ "$1" == "bump:"* ]]
         | grep new_version | sed -r s,"^.*=",,`
 
     echo "Bumping to version $new_version"
-    git push
+    git tag -a $new_version -m "Release $new_version"
```
PierreLeveau, line 36: If we do that, we add the tag before the PR is actually merged to master. If I remember well the tag will be lost when merging.

theodu: Ah yes right, I remove it. 
My other idea would be to create a workflow to add the tag and create the draft release whenever a branch that begins by "release/" is merged. What do you think ?

PierreLeveau (approved):

entrypoint.sh
PierreLeveau: that could be great! the draft release link could be sent in the #python-sdk channel.

PR #686, opened 2022-09-29
FIX tqdm and Frame import

theodu:
- fix: fix signed_urls request
- fix: tqdm used

### DIFF
--- a/kili/utils/bucket.py
+++ b/kili/utils/bucket.py
@@ -5,16 +5,17 @@ from typing import Union
 from urllib.parse import parse_qs, urlparse
 
 import requests
+from kili.authentication import KiliAuth
+from kili.graphql.operations.asset.queries import GQL_CREATE_UPLOAD_BUCKET_SIGNED_URLS
 from tenacity import retry
 from tenacity.stop import stop_after_attempt
 from tenacity.wait import wait_random
 
-from kili.authentication import KiliAuth
-from kili.graphql.operations.asset.queries import GQL_CREATE_UPLOAD_BUCKET_SIGNED_URLS
-
 AZURE_STRING = "blob.core.windows.net"
 GCP_STRING = "storage.googleapis.com"
 GCP_STRING_PUBLIC = "storage.cloud.google.com"
+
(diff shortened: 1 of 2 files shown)


kili/services/asset_import/base.py
```diff
@@ -14,9 +14,10 @@
 )
 from kili.helpers import format_result, is_url
 from kili.orm import Asset
+from tqdm import tqdm
+
 from kili.queries.asset import QueriesAsset
 from kili.utils import bucket, pagination
-from kili.utils.tqdm import tqdm
```
PierreLeveau, line 19: you should rather update `kili/utils/tqdm.py` with `tqdm import tqdm` to keep things centralized.

kili/utils/bucket.py
```diff
@@ -5,17 +5,18 @@
 from urllib.parse import parse_qs, urlparse
 
 import requests
+from kili.authentication import KiliAuth
+from kili.graphql.operations.asset.queries import GQL_CREATE_UPLOAD_BUCKET_SIGNED_URLS
```
PierreLeveau: kili import should not be there. This is probably because there has been a `pip install kili` in your environment

kili/utils/bucket.py
```diff
@@ -25,12 +26,16 @@ def request_signed_urls(auth: KiliAuth, project_id: str, size: int):
...
     """
-    payload = {
-        "projectID": project_id,
-        "size": size,
-    }
-    urls_response = auth.client.execute(GQL_CREATE_UPLOAD_BUCKET_SIGNED_URLS, payload)
-    return urls_response["data"]["urls"]
+    signed_urls = []
+    while len(signed_urls) < size:
+        nb_url_requested = min(MAX_NUMBER_SIGNED_URLS_TO_FETCH, size - len(signed_urls))
```
PierreLeveau, line 31: I think we should let it crash. This is the caller of `request_signed_urls` that should make sure there not too many rsl requested.

PierreLeveau (changes requested):

PierreLeveau (approved):

PR #689, opened 2022-09-29
AAU, on the SDK, when I look at the assets method, I have the possibility to download content in local

theodu:
- fix: flaky test
- feat: implement asset download


kili/services/asset_download/exceptions.py
```diff
@@ -0,0 +1,7 @@
+"""
+Asset download service exceptions
+"""
+
+
+class MissingPropertyError(Exception):
```
PierreLeveau, line 6: You can inherit from [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) to be able to pass a message.

kili/services/asset_download/exceptions.py
```diff
@@ -0,0 +1,7 @@
...
+    """Raised when trying to download an asset media with a missing fiels"""
```
PierreLeveau, line 7: fields

PierreLeveau (changes requested):

kili/services/__init__.py
```diff
@@ -2,8 +2,9 @@
 Python SDK service layer
 """
 
+from kili.services.asset_download import download_asset_media
```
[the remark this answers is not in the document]
theodu: I like the get_asset service idea because we will implement other processing later. For example the get_assets features in autoML should be integrated into the SDK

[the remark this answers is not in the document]
PierreLeveau: 👍

kili/queries/asset/exceptions.py
```diff
@@ -0,0 +1,7 @@
+"""
+Asset queries related exceptions
+"""
+
+
+class MissingPropertyError(Exception):
+    """Raised when trying to download an asset media with a missing fiels"""
```
PierreLeveau, line 7: field

kili/queries/asset/helpers.py
```diff
@@ -0,0 +1,94 @@
...
+                " querying assets "
+            )
+        field_array.append(value)
+    return field_array
+
+
+def download_asset_media(assets: List[dict], local_dir_path: Path) -> List[dict]:
+    """Download assets media in local"""
+    local_dir_path.mkdir(parents=True, exist_ok=True)
+    content_array: List[str] = get_field_array(assets, "content")
```
PierreLeveau, line 88: Note on the perf, which is not mandatory: this will require to loop over the asset several times, which can be cumbersome with large asset list. You a can return a list of arrays, and unwrap it into content_array and content_id_array.

kili/queries/asset/helpers.py
```diff
@@ -0,0 +1,94 @@
...
+    external_id_array: List[str] = get_field_array(assets, "externalId")
+    with ThreadPoolExecutor() as threads:
+        path_gen = threads.map(
+            download_file, content_array, external_id_array, repeat(local_dir_path)
```
PierreLeveau, line 92: I did not now `repeat`, this is nice

kili/queries/asset/helpers.py
PierreLeveau: and this could be generators too, to spare some memory in the same cases.

PierreLeveau (approved):
Some suggestions but LGTM

kili/queries/asset/helpers.py
theodu: yes right! I replaced it with generators. For the perf issue, I wait to see what we do for the json_content because it is a field that we might not query every time

PR #695, opened 2022-10-04
AAU, on the SDK, when I look at the SDK V2, I see a function to export a project

theodu:


kili/project/__init__.py
```diff
@@ -27,8 +27,8 @@ def __init__(  # pylint: disable=too-many-arguments
     @typechecked
     def export(  # pylint: disable=too-many-arguments
         self,
-        output_filename: str,
-        output_format: LabelFormat,
+        filename: str,
+        fmt: LabelFormat,
```
theodu, line 31: These arguments appear to me to be easier to use because shorter, without making any confusion for the end user. 
Could be discussed if you disagree.
I used this on the SDK V2 function and updated the one on the SDK V3

kili/queries/label/__init__.py
```diff
@@ -314,3 +318,58 @@ def count_labels(
         result = self.auth.client.execute(GQL_LABELS_COUNT, variables)
         count = format_result("data", result)
         return int(count)  # type:ignore
+
+    def export_labels(
```
theodu, line 322: I put it in the QueriesLabel because there is already the `export_labels_as_df function`. Does not have any impact when using on python it but it changes where the function reference is in the documentation. The other possibility would be in the QueriesProject.

kili/project/__init__.py
PierreLeveau: 👍

kili/queries/label/__init__.py
PierreLeveau: it's fine to put it here, to remain consistent with V2.

PierreLeveau (approved):

PR #702, opened 2022-10-06
AAU, on SDK repo, when I look at the architecture, I see a src folder

theodu:


PierreLeveau (changes requested):
👍 small changes but good!

PR #705, opened 2022-10-07
feat: add multiple_files argument to split labels into multiple files

antonioloison:
Add `--multiple-files` argument to split labels into several files in export


kili/cli/project/export.py
```diff
@@ -21,14 +21,21 @@
...
+)
 @click.option(
     "--layout",
     type=click.Choice(get_args(SplitOption)),
     default="merged",
     help="Layout of the label files",
 )
 @click.option(
-    "--output-file", type=str, help="File into which the labels are saved.", required=True
+    "--multiple-files",
```
PierreLeveau, line 34: Even if it is not what has been done for the Kili raw format, the default should be multiple files. This is what is done for most of the image annotations formats in Kili. 
So we should set --single-file as a flag, defaulting to False. But I'm interested in you opinion about that!

PierreLeveau (changes requested):

kili/cli/project/export.py
antonioloison: I am ok with that. I think we shoud be consistent with the other exports

src/kili/services/export/format/kili/common.py
```diff
@@ -84,9 +84,20 @@ def _save_assets_export(self, assets: List[Dict], output_filename: str):
...
+                with open(os.path.join(base_folder, "data.json"), "wb") as output_file:
+                    output_file.write(project_json.encode("utf-8"))
+            else:
+                labels_folder = os.path.join(base_folder, "labels")
+                os.makedirs(labels_folder)
+                for asset in assets:
+                    external_id = asset["externalId"].replace(" ", "_")
+                    asset_json = json.dumps(asset, sort_keys=True, indent=4)
+                    with open(
+                        os.path.join(labels_folder, f"asset_{external_id}.json"), "wb"
```
theodu, line 98: when importing labels on an asset called 'foo' from the CLI, we search for the file called 'foo.json' so maybe we should rather call the exported file 'foo.json' ?
Also the prefix asset is disturbing because it is not an asset but a label.

theodu (changes requested):

src/kili/services/export/format/kili/common.py
PierreLeveau: 👍

PierreLeveau (approved):

theodu (approved):

PR #712, opened 2022-10-14
chore: check that the PR title respects commit conventions

PierreLeveau:
image


theodu:
so this only triggers on a pull request ?
What happen if I made 3 commits in local, 2 that are compatible and one not and that I then push the branch and create a pull request ?

theodu (approved):

PR #718, opened 2022-10-21
docs: added details about label export

PierreLeveau:
Here is the new page dedicated to label export. @xczuba we may also add this to the Kili documentation.


theodu (changes requested):
if label_export is not set in the nav in the mkdocs.yaml file, it will not appear on the website.
To launch the website in local to test it, you can run `mike serve`

PierreLeveau:
> if label_export is not set in the nav in the mkdocs.yaml file, it will not appear on the website. To launch the website in local to test it, you can run `mike serve`
I don't think it should go into the reference documentation, which lists the entities and their methods. So I put it as a separate README for now. Exporting probably deserves a dedicated age, and/or a tutorial like in Python SDK tutorials (like in https://docs.kili-technology.com/docs/python-sdk-tutorials) or a recipe (https://docs.kili-technology.com/recipes).

theodu (approved):
Ok I see then!

PR #727, opened 2022-10-25
feat: disable tqdm when appending many assets to dataset

obendidi:


PierreLeveau (approved):

PierreLeveau:
thanks for the contribution @obendidi !

PR #730, opened 2022-10-26
feat: throw error when export format is incompatible with format

antonioloison:
- Throw format value when using single file argument
- Add ValueError raise test


src/kili/services/export/__init__.py
```diff
@@ -34,6 +34,9 @@ def export_labels(  # pylint: disable=too-many-arguments
     if kili.count_projects(project_id=project_id) == 0:
         raise NotFound(f"project ID: {project_id}")
 
+    if single_file and label_format != "raw":
+        raise ValueError(f"The label format {label_format} can not be exported in a single file.")
```
PierreLeveau, line 38: you can move this verification deeper in the code (in `KiliFromatter`), it is the responsibility of the Exporter to tell if he can handle option combinations or not.

antonioloison: done, I moved it to the BaseExporter

PR #737, opened 2022-10-28
feat(cli): add COCO export

PierreLeveau:
- Incomplete coco
- wip


src/kili/queries/label/__init__.py
```diff
@@ -331,6 +331,7 @@ def export_labels(
             layout: Layout of the exported files: "split" means there is one folder
                 per job, "merged" that there is one folder with every labels.
             disable_tqdm: Disable the progress bar if True.
+            download_media: Tell to download media if True.
```
theodu, line 334: `download_media` is not an arg of the function

src/kili/services/export/format/base.py
```diff
@@ -145,9 +148,69 @@ def export_project(
...
+            download_media=self.download_media,
         )
         self.process_and_save(assets, export_params.output_file)
 
+    @staticmethod
+    def _filter_out_autosave_labels(assets: List[Asset]) -> List[Asset]:
+        """
+        Removes AUTOSAVE labels from exports
+
+        Parameters
```
theodu, line 160: numpy doctrings style

src/kili/services/export/format/coco/common.py
```diff
@@ -0,0 +1,273 @@
...
+    images: List[_CocoImage]
+    annotations: List[_CocoAnnotation]
+
+
+class CocoExporter(BaseExporter):
+    """
+    Common code for Kili exporters.
+    """
+
+    download_media = True
```
theodu, line 65: why do you need to download the asset here ?

src/kili/services/export/types.py
```diff
@@ -21,3 +21,29 @@ class JobCategory(NamedTuple):
 
 
 YoloAnnotation = Tuple[int, float, float, float, float]
+
+
+InputType = Literal["IMAGE", "TEXT"]
+MLTask = Literal["CLASSIFICATION", "NAMED_ENTITIES_RECOGNITION", "OBJECT_DETECTION"]
+Tool = Literal["rectangle", "semantic", "polygon"]
+JobName = NewType("JobName", str)
+ProjectId = NewType("ProjectId", str)
```
theodu, line 30: We add a lot of duplicate types like these. We should put the most frequent one like these at the root of the package

src/kili/services/export/format/coco/common.py
PierreLeveau: COCO is self-contained and requires paths to the media files
(comment added)

src/kili/services/export/types.py
PierreLeveau: moved to `services/types.py`

src/kili/services/export/types.py
```diff
@@ -21,3 +21,29 @@ class JobCategory(NamedTuple):
...
+    isChild: bool
+    tools: List[Tool]
+    mlTask: MLTask
+    models: Any  # example: {"interactive-segmentation": {"job": "SEMANTIC_JOB_MARKER"}},
+    isVisible: bool
+    required: int
+    isNew: bool
+
+
+Jobs = Dict[JobName, Job]
```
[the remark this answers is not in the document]
PierreLeveau: moved to `services/types.py`

theodu (changes requested):
error when executing the command

theodu (approved):

PR #738, opened 2022-10-31
feat: append new function append_labels to append several labels at the same time

theodu:


src/kili/mutations/label/__init__.py
```diff
@@ -102,15 +103,20 @@ def append_to_labels(
         project_id: Optional[str] = None,
         seconds_to_label: Optional[int] = 0,
     ):
-        """Append a label to an asset.
+        """
+        !!! danger "[DEPRECATED]"
```
PierreLeveau, line 107: shouldn't we use the deprecated decorator?

src/kili/mutations/label/__init__.py
```diff
@@ -142,6 +154,65 @@ def append_to_labels(
...
 
+    @typechecked
+    def append_labels(
+        self,
+        json_response_array: List[dict],
+        author_id_array: Optional[List[str]] = None,
+        label_asset_external_id_array: Optional[List[str]] = None,
+        label_asset_id_array: Optional[List[str]] = None,
+        seconds_to_label_array: Optional[List[int]] = None,
+        label_type: str = "DEFAULT",
```
PierreLeveau, line 165: maybe the time to use a Literal type for label_type

tests/test_client.py
```diff
@@ -6,8 +6,8 @@
 
 import pytest
 
-from kili.client import Kili
-from kili.exceptions import AuthenticationFailed
+from src.kili.client import Kili
```
PierreLeveau, line 9: this should not happen, src is not a package

PierreLeveau (changes requested):

src/kili/mutations/label/__init__.py
```diff
@@ -92,6 +94,12 @@ def generate_variables(batch):
         )
         return format_result("data", results[0], Label)
 
+    @deprecate(
+        msg=(
+            "append_to_labels is deprecated. Please use append_labels instead. This new function"
```
PierreLeveau, line 99: method

src/kili/mutations/label/__init__.py
```diff
@@ -154,61 +155,49 @@ def append_to_labels(
...
-        json_response_array: List[dict],
-        author_id_array: Optional[List[str]] = None,
-        label_asset_external_id_array: Optional[List[str]] = None,
-        label_asset_id_array: Optional[List[str]] = None,
-        seconds_to_label_array: Optional[List[int]] = None,
-        label_type: str = "DEFAULT",
-        project_id: Optional[str] = None,
-    ):
-        """Append a label to an asset.
+    def append_labels(self, labels: List[AppendLabelData], label_type: LabelType = "DEFAULT"):
```
PierreLeveau, line 158: should be typechecked imho (this is what we do for public methods)

PierreLeveau: we should not force the user to use a TypeDict. So we should have
```suggestion
    def append_labels(self, labels: List[Dict], label_type: LabelType = "DEFAULT"):
```

src/kili/mutations/label/__init__.py
```diff
@@ -154,61 +155,49 @@ def append_to_labels(
...
         Returns:
             A result object which indicates if the mutation was successful,
                 or an error message.
 
         Examples:
             >>> kili.append_to_labels(
-                label_asset_id_array=[asset_id1, asset_id2],
-                json_response=[{...}, {...}]
+                    [
+                        {'json_response': {...}, asset_id: 'cl9wmlkuc00050qsz6ut39g8h'},
```
PierreLeveau, line 182: This is a "V3"-style syntax, are we sure we should push it? No pb to use it behind the API, but imho we should remain consistent with the other methods.

PierreLeveau (changes requested):

src/kili/mutations/label/__init__.py
theodu: Even the resolver in the backend now has the style of array of dict instead of several arrays of fields. So if people want to use the resolver directly in graphQL, they will have to send the labels as array of dict. It is indeed not in line with other methods but for me we should then begin to switch, don't you think ? No conversion needed in the function and easier to use for the user.

src/kili/services/label_import/__init__.py
```diff
@@ -63,3 +68,30 @@ def import_labels_from_files(  # pylint: disable=too-many-arguments
...
+    ]
+    batch_generator = pagination.batch_iterator_builder(labels_data)
+    result = []
+    for batch_labels in batch_generator:
+        variables = {
+            "data": {"labelType": label_type, "labelsData": batch_labels},
+            "where": {"idIn": [label["assetID"] for label in batch_labels]},
+        }
+        batch_result = kili.auth.client.execute(GQL_APPEND_MANY_LABELS, variables)
+        result.extend(format_result("data", batch_result, Label))
```
PierreLeveau, line 96: `List[Label]` ? not sure

PierreLeveau (approved):

src/kili/services/label_import/__init__.py
theodu: no, the array feature is handled in the format_result already is exploited in the format_result

theodu:
@PierreLeveau ready to review. With recipe changes but not with CLI changes

PierreLeveau (approved):
LGTM. Did you launch the the e2e tests?

PR #748, opened 2022-11-08
fix(COM-325): stringify json contents when importing tiff data

mDuval1:


theodu (approved):

PierreLeveau (approved):
Can this wait tomorrow's release?

PierreLeveau (changes requested):
we should not release at this stage, otherwise this patch will embed all the features of the release planned tomorrow

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.123.2
+current_version = 2.123.3
```
PierreLeveau, line 2: do not bump version

src/kili/__init__.py
```diff
@@ -5,4 +5,4 @@
 Kili Python SDK
 """
 
-__version__ = "2.123.2"
+__version__ = "2.123.3"
```
PierreLeveau, line 8: do not bump version

PierreLeveau (changes requested):

PierreLeveau (approved):

PR #752, opened 2022-11-08
AA Developer U, on upload_plugin, when the method fail because of duplicates name, I see a hint to use update_plugin

florianlega:
✅ Closes: DATA-807


src/kili/services/plugins/tools.py
```diff
@@ -0,0 +1,13 @@
+"""
```
pmKili, line 1: Why calling this file tools and not helpers?

src/kili/services/plugins/tools.py
```diff
@@ -0,0 +1,13 @@
...
+
+DUPLICATE_ERROR_SUBSTRING = "an entity Plugin already exists with value"
+
+def check_errors_plugin_upload(result, file_path, plugin_name):
+    """
+    Check if the error during upload is because the plugin already
+    exists and print the upload function for the user
+    """
+    if "errors" in result and DUPLICATE_ERROR_SUBSTRING in result["errors"][0]["message"]:
+        print('Perhaps you wanted to use update_plugin("%s", "%s") (Warning, the existing plugin will be discarded)' % (file_path, plugin_name))
```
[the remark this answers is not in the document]
pmKili: And you should use the logger as @gheorghetutunaru did in his PR with fstrings

src/kili/services/plugins/tools.py
florianlega: I wanted to keep same wording as the export service

src/kili/services/plugins/tools.py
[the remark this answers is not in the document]
gheorghetutunaru: If we see that there is this error, shouldn't we throw an Exception instead of just logging ? Because now we will have this log, but also all the graphql error being printed, which is kind of redundant

src/kili/services/plugins/tools.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def check_errors_plugin_upload(result, file_path, plugin_name):
+    """
+    Check if the error during upload is because the plugin already
+    exists and print the upload function for the user
+    """
+    if "errors" in result and DUPLICATE_ERROR_SUBSTRING in result["errors"][0]["message"]:
+        print(
+            f'Perhaps you wanted to use update_plugin("{file_path}", "{plugin_name}") '
```
p-desaintchamas, line 15:
```suggestion
            f'Perhaps you wanted to use kili.update_plugin(file_path="{file_path}", plugin_name="{plugin_name}") '
```
And perhaps handle the case where one of those variables is not defined ?

florianlega: Only plugin_name can be undefined but the method that calls this function set it as the file name so no need to check it

PR #761, opened 2022-11-10
fix(import): fix pyright pipeline

pmKili:


src/kili/services/asset_import/text.py
```diff
@@ -91,10 +91,13 @@ def import_assets(self, assets: List[AssetLike]):
                 self.auth, self.project_params, batch_params, self.pbar
             )
         elif data_type == TextDataType.RAW_TEXT:
-            assets: List[AssetLike] = [
-                AssetLike(**{**asset, "content": asset["content"].encode("utf-8")})
-                for asset in assets
-            ]
+            for i, asset in enumerate(assets):
```
PierreLeveau, line 94:
```
for asset in assets:
    if "content" in asset:
         asset["content"] = asset["content"].encode("utf-8")
```
should work because `asset` are references to dict (changes will be done in place)

PierreLeveau (changes requested):

src/kili/services/asset_import/text.py
pmKili: yes it works but I must add a `# type: ignore` because `asset['content']` is a str and `asset["content"].encode("utf-8")` is bytes. Is it ok for you?

pmKili: Otherwise, I can update the type of `content` in `AssetLike` to `content: Union[str, bytes]`

PierreLeveau: I'm in favor of putting `content: Union[str, bytes]`, the type error is relevant

PierreLeveau (approved):

PR #787, opened 2022-11-21
fix: add test for import of video legacy from content_array urls

theodu:


PierreLeveau (approved):
👍 do we have a e2e test / notebook that imports into a legacy_project as well ?

theodu:
> 

As it is legacy, I don't think that we have any recipe/notebook that uses this. I will ask to the product team what is the status of legacy video

PierreLeveau (approved):

PR #808, opened 2022-12-02
fix: add missing type hints

Jonas1312:


src/kili/graphql_client.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         self.token = None
         self.verify = verify
 
-    def execute(self, query, variables=None):
+    def execute(self, query, variables=None) -> dict:
```
PierreLeveau, line 45: `dict`-> `Dict` , here and below. We use Python types and not the native structures for typing, to stay consistent.

PierreLeveau: (it may be not the case everywhere in the existing code, but that's a mistake)

src/kili/subscriptions/label/__init__.py
```diff
@@ -34,7 +36,7 @@ def label_created_or_updated(self, project_id: str, callback: Callable[[str, str
             callback: This function takes as input the id of the asset and its content.
 
         Returns:
-            A subscription client
+            SubscriptionGraphQLClient: A subscription client.
```
PierreLeveau, line 39: no need to specify the return type here

PierreLeveau (changes requested):

PierreLeveau (approved):

PR #813, opened 2022-12-06
ci: create minor release branch workflow

Jonas1312:
CI: https://github.com/kili-technology/kili-python-sdk/actions/runs/3646440104/jobs/6157525891

Created branch: https://github.com/kili-technology/kili-python-sdk/tree/release/2.127.0

remember to delete release/2.127.0 branch after merging

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -33,7 +33,7 @@ function create_release_branch() {
 
     #get the new version
     new_version=$(bump_version dry-run)
-    echo "New version: $new_version"
+    echo "New version (bump_version dry-run): $new_version"
 
     #create a branch from the specified sha and commit the version bump
     git config user.name
@@ -41,7 +41,9 @@ function create_release_branch() {
 
     git checkout -B release/$new_version $commit
     new_version=$(bump_version commit)
-    if git push --force --quiet --set-upstream origin; then
+    echo "New version (bump_version commit): $new_version"
+
+    if git push --force --quiet --set-upstream origin release/$new_version $commit; then
         echo "version bump commited and pushed on the release branch"
         echo "Tests are launched on Github Actions: https://github.com/kili-technology/kili-python-sdk/actions"
     else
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -13,26 +13,35 @@ function bump_version(){
 }
 
 function create_release_branch() {
-    read -p 'Bump type (possible values: patch or minor): ' release_type
+    release_type=$1
+
```
PierreLeveau, line 17: you can add a check on the argument number then

entrypoint.sh
```diff
@@ -13,26 +13,35 @@ function bump_version(){
...
+    echo "New version (bump_version dry-run): $new_version"
 
-    #create a branch from the specified sha and and commit the version bump
+    #create a branch from the specified sha and commit the version bump
     git checkout -B release/$new_version $commit
     new_version=$(bump_version commit)
-    if git push -f -q; then
+    echo "New version (bump_version commit): $new_version"
+
+    if git push --force --quiet --set-upstream origin release/$new_version; then
```
PierreLeveau, line 39: I am not sure there is a good reason to push force after all, you can remove it.

PierreLeveau (changes requested):

entrypoint.sh
Jonas1312: Well, the check is already done, I think.

There are two arguments for this function.

One mandatory for the bump type, it is checked here: https://github.com/kili-technology/kili-python-sdk/blob/2c8abb4d315cb629eeb9494a7b8798d06e2ad3f4/entrypoint.sh#L18-L21  If the bump type argument is not provided and is not equal to `minor` or `patch`, it will exit anyway.

And one optional argument for the commit hash, that will be set to HEAD by default: https://github.com/kili-technology/kili-python-sdk/blob/2c8abb4d315cb629eeb9494a7b8798d06e2ad3f4/entrypoint.sh#L25

I think however I should add some comments in the script about how to use it.

PierreLeveau (approved):

PR #819, opened 2022-12-09
chore: remove the api key mutations

PierreLeveau:


theodu (changes requested):
We should also remove the api_key query from the clients functions
and  you missed to remove it from the doc

theodu (approved):

PR #821, opened 2022-12-12
refactor: move project_user mutations

PierreLeveau:
preliminary refactoring


theodu (approved):
Good! I didn't notice this.
Putting the whole mutation string in the function is not very clean but ok because we will change it soon

PierreLeveau:
> Good! I didn't notice this. Putting the whole mutation string in the function is not very clean but ok because we will change it soon

It not that dirty: the string is only used once, and that does not make the code less readable. Anyway we will decide the ost convenient way to do it soon.

PR #830, opened 2022-12-14
fix: get_content_stream remove headers

Jonas1312:


theodu (approved):

PierreLeveau (changes requested):
the test should be moved to `tests/integration` as it requires a "real" Kili client connected to a Kili platform.
Btw `tests/integration` should be named `tests/e2e`, can be the right time to change the name (and in the CI workflows as well)

PierreLeveau (approved):

PR #832, opened 2022-12-15
chore: remove project_user mutation

PierreLeveau:


theodu (changes requested):
I just checked in the kili app codebase and saw that this function is used in the backend integration tests. I am not sure that we use an old version of kili in the tests. I see 2.123.2 on the tests.
We should rather only remove it from the doc I think

theodu (approved):

PR #838, opened 2022-12-20
feat: add option to get assets in export project

Jonas1312:


src/kili/services/export/format/base.py
```diff
@@ -146,6 +147,10 @@ def export_project(
         """
         self._check_arguments_compatibility()
         self.logger.warning("Fetching assets...")
+        if self.download_media:
+            count = self.kili.count_assets(project_id=self.project_id)
+            if count > self.THRESHOLD_WARN_MANY_ASSETS:
+                self.logger.warning(f"Downloading {count} assets. This might take a while.")
```
PierreLeveau, line 153: imho it would be worth if to put this verification in the "client front-end layer", and add a front-end specific message on how to disable the downloading.
200 is low imho, you can put 1000 (even if this is content size specific of course)
Also, you can use `warnings.warn` to let the user specify if he/she wants to disable the specified warning.

src/kili/services/export/format/coco/__init__.py
```diff
@@ -58,10 +58,12 @@ class _CocoFormat(TypedDict):
...
 class CocoExporter(AbstractExporter):
     """
-    Common code for Kili exporters.
+    Common code for COCO exporter.
     """
 
-    download_media = True  # COCO is self-contained and requires paths to the media files
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.download_media = True  # we need the assets for COCO annotation format
```
PierreLeveau, line 66: if this differs from the `download_media` arg is false, you can output a warning.

src/kili/services/export/format/kili/__init__.py
```diff
@@ -43,6 +45,15 @@ def _save_assets_export(self, assets: List[Asset], output_filename: Path):
                     with (labels_folder / f"{external_id}.json").open("wb") as output_file:
                         output_file.write(asset_json.encode("utf-8"))
             self.create_readme_kili_file(root_folder)
+
+            if self.download_media:
+                images_folder = base_folder / "images"
+                images_folder.mkdir(parents=True)
+                for asset in assets:
+                    img_path = images_folder / f'{asset["externalId"]}.jpg'
```
PierreLeveau, line 53: extension should depend on the original content type. You can piggy back on the extension of the local file, for example with pattern matching `f"{external_id}.*"`

src/kili/services/export/format/voc/__init__.py
```diff
@@ -22,6 +23,10 @@ class VocExporter(AbstractExporter):
     Common code for VOC exporter.
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.download_media = True  # we need the assets for Pascal VOC annotation format
```
PierreLeveau, line 28: same as COCO, output a warning.

src/kili/services/export/format/yolo/__init__.py
```diff
@@ -318,6 +326,17 @@ def _process_asset(
         _write_labels_to_file(labels_folder, filename, frame_labels)
 
         content_frame = content_frames[idx] if content_frames else asset["content"]
+
+        img_path = images_folder / f"{filename}.jpg"
```
PierreLeveau, line 330: same here, we should not hardcode the file extension

PierreLeveau (changes requested):

src/kili/services/export/format/base.py
Jonas1312: in what module would you put the verification exactly?

src/kili/services/export/format/coco/__init__.py
```diff
@@ -63,7 +64,9 @@ class CocoExporter(AbstractExporter):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.download_media = True  # we need the assets for COCO annotation format
+        if not self.download_media:
+            warnings.warn("For COCO export, download_media must be set to True.")
```
PierreLeveau, line 68: `"For the export to the COCO format, the download of assets cannot be disabled"`
`download_media` is not the parameter name in the CLI.

src/kili/services/export/format/voc/__init__.py
```diff
@@ -25,7 +26,9 @@ class VocExporter(AbstractExporter):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.download_media = True  # we need the assets for Pascal VOC annotation format
+        if not self.download_media:
+            warnings.warn("For Pascal VOC export, download_media must be set to True.")
```
PierreLeveau, line 30: "For the export to the Pascal VOC format, the download of assets cannot be disabled"
download_media is not the parameter name in the CLI.

src/kili/services/export/format/kili/__init__.py
```diff
@@ -50,9 +50,8 @@ def _save_assets_export(self, assets: List[Asset], output_filename: Path):
                 images_folder = base_folder / "images"
                 images_folder.mkdir(parents=True)
                 for asset in assets:
-                    img_path = images_folder / f'{asset["externalId"]}.jpg'
                     if os.path.isfile(asset["content"]):
-                        shutil.move(asset["content"], img_path)
+                        shutil.move(asset["content"], images_folder)
```
PierreLeveau, line 54: I did not see it the first time, but we should do a copy since we do not want to remove the files if they have been downloaded through a user call to `.assets`.
The nicest way to do that would be specify the download folder directly in the `.assets` call in the export implementation.

PierreLeveau (changes requested):

src/kili/project/__init__.py
```diff
@@ -35,6 +35,7 @@ def export(  # pylint: disable=too-many-arguments
         single_file: bool = False,
         disable_tqdm: bool = False,
         log_level: LogLevel = "INFO",
+        download_media: bool = True,
```
theodu, line 38: I would call it `with_assets` too or `download_assets`
In the `assets` method, we were directly refering to assets, here we are dealing with a project. I find it more natural to speak about assets and not media. 
Plus, for me the name should be the same as in the CLI to avoid confusions

src/kili/project/__init__.py
```diff
@@ -57,6 +58,7 @@ def export(  # pylint: disable=too-many-arguments
                 only available for some specific formats (COCO and Kili).
             disable_tqdm: Disable the progress bar if True.
             log_level: Level of debugging.
+            download_media: Download the assets in the export.
```
theodu, line 61: same here

src/kili/queries/label/__init__.py
```diff
@@ -319,6 +319,7 @@ def export_labels(
         layout: SplitOption = "split",
         single_file: bool = False,
         disable_tqdm: bool = False,
+        download_media: bool = True,
```
theodu, line 322: same here

src/kili/services/export/format/kili/__init__.py
```diff
@@ -43,6 +45,14 @@ def _save_assets_export(self, assets: List[Asset], output_filename: Path):
...
                         output_file.write(asset_json.encode("utf-8"))
             self.create_readme_kili_file(root_folder)
+
+            if self.download_media:
+                images_folder = base_folder / "images"
+                images_folder.mkdir(parents=True)
+                for asset in assets:
+                    if os.path.isfile(asset["content"]):
+                        shutil.move(asset["content"], images_folder)
+
```
theodu, line 55: can't you find a way to define the base folder before to direclty download the assets in the right folder when calling assets ?

src/kili/services/export/format/voc/__init__.py
```diff
@@ -74,42 +70,42 @@ def _process_asset_for_job(
...
-    for idx, frame in frames.items():
-        if is_frame:
-            filename = f"{asset['externalId']}_{str(idx + 1).zfill(leading_zeros)}"
-            video_filenames.append(filename)
-        else:
-            filename = asset["externalId"]
+    filename = asset["externalId"]
+    img_path = images_folder / Path(asset["content"]).name
+    if not os.path.isfile(content_frame):
+        download_file(content_repository, content_frame, img_path)
```
theodu, line 102: In the assets query, we download assets of a batch just after having query that batch. The reason is that we use signed URLS that have a limit valid time. if you have a dataset of 20 000 assets, that you query all assets and then download them all, there is high chance that the last signed URLS will not be valid anymore!
You should download assets with the assets method and process them

theodu: And you already downloaded them in the export_project function, in which case would they not be downloaded in order to have to redownload them here ?

src/kili/services/export/format/yolo/__init__.py
```diff
@@ -318,11 +326,22 @@ def _process_asset(
...
         content_frame = content_frames[idx] if content_frames else asset["content"]
+
+        img_path = images_folder / Path(asset["content"]).name
+        if download_media and not os.path.isfile(img_path):
+            if not os.path.isfile(content_frame):
+                _write_content_frame_to_file(
+                    content_frame, images_folder, Path(asset["content"]).name, content_repository
+                )
+            else:
+                shutil.move(content_frame, img_path)
```
theodu, line 337: same remark for the move here

src/kili/project/__init__.py
Jonas1312: I agree with you that `download_assets` makes more sense, but we use `download_media` a lot in the repository. wdyt @PierreLeveau

src/kili/services/export/format/yolo/__init__.py
theodu: and for the download

theodu (changes requested):
questions on assets download and parameter names

src/kili/services/export/format/base.py
PierreLeveau: in `kili.export_labels` and `kili.cli.project.export`

src/kili/services/export/format/kili/__init__.py
PierreLeveau: 👍

src/kili/project/__init__.py
PierreLeveau: I agree with @theodu 's point, `download_assets` is more relevant in the export use case.

src/kili/services/export/__init__.py
```diff
@@ -28,6 +28,7 @@ def export_labels(  # pylint: disable=too-many-arguments, too-many-locals
     output_file: str,
     disable_tqdm: bool,
     log_level: LogLevel,
+    download_media: bool,
```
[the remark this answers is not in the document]
PierreLeveau: `with_assets` is fine

src/kili/services/export/format/base.py
Jonas1312: I put the warning in `services.export_labels` since this method is called from several places (CLI, QueriesLabel and Project). It's easier and better to keep it in one single place, I think.

Jonas1312:
- I renamed everything to `with_assets`
- In the `AbstractExporter`, a temporary root export directory is created so that we can download the assets directly into it with `fetch_assets`
- I removed the video export in Pascal VOC since it hasn't been tested and makes the code harder to read/maintain
- I had to remove the Pascal VOC CLI test since the mocked `asset["content"]` is not downloadable

theodu (approved):

tests/services/export/test_export.py
```diff
@@ -101,19 +145,52 @@ def get_file_tree(folder: str):
...
+        (
+            "Export to Kili raw format",
+            {
+                "export_kwargs": {
+                    "project_id": "object_detection",
+                    "label_format": "raw",
+                    "with_assets": True,
+                },
+                "file_tree_expected": {
+                    "images": {"car_1.jpg": {}},
```
PierreLeveau, line 188: the folder should be named "assets"

PierreLeveau: because the content is not necessarily images

PierreLeveau (changes requested):

tests/services/export/test_export.py
PierreLeveau: (for the Kili format only)

PierreLeveau (approved):

PR #839, opened 2022-12-21
docs: add more info about search_query parameter

Jonas1312:


src/kili/queries/project/__init__.py
```diff
@@ -56,6 +56,7 @@ def projects(
         Args:
             project_id: Select a specific project through its project_id.
             search_query: Returned projects with a title or a description matching this string.
+                The query must follow the PostgreSQL ILIKE operator [specification](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE).
```
theodu, line 59: I find it quite disturbing because just writing search_query='project test' works.
By reading this, I would understand that I have to write something like 'project.title LIKE project test'

theodu (changes requested):

PierreLeveau (approved):

theodu (approved):

theodu (approved):

PR #850, opened 2022-12-28
chore: abstract queries in a graphql folder

theodu:


src/kili/cli/project/member/add.py
```diff
@@ -72,13 +76,19 @@ def add_member(
...
+    existing_members = ProjectUserQuery(kili.auth.client)(
+        where=ProjectUserWhere(project_id=project_id),
+        fields=[
+            "activated",
+            "user.email",
+        ],
+    )
+    existing_members_email = [
         member["user"]["email"] for member in existing_members if member["activated"]
     ]
```
Jonas1312, line 88: a set() could be a bit faster for projects with many members

src/kili/cli/project/member/helpers.py
```diff
@@ -49,31 +53,32 @@ def collect_members_from_csv(csv_path: str, role: Optional[str]):
...
         )
-        for user in users:
-            if user["activated"]:
-                members.append({"email": user["user"]["email"], "role": user["role"]})
+        for existing_member in existing_members:
+            if existing_member["activated"]:
+                activated_members.append(
+                    {"email": existing_member["user"]["email"], "role": existing_member["role"]}
+                )
     except:
```
Jonas1312, line 74: we don't have a specific error for this except?

src/kili/cli/project/member/add.py
```diff
@@ -72,13 +76,19 @@ def add_member(
...
-    existing_members = kili.project_users(project_id=project_id, disable_tqdm=True)
-    existing_members = [
+    existing_members = ProjectUserQuery(kili.auth.client)(
+        where=ProjectUserWhere(project_id=project_id),
+        fields=[
+            "activated",
+            "user.email",
+        ],
+    )
+    existing_members_email = [
```
PierreLeveau, line 86: `existing_member_emails`

src/kili/graphql/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+
+from abc import ABC, abstractmethod
+from typing import Any, Dict, Iterable, List, NamedTuple, Optional, Type
+
+from tqdm import tqdm
+
+from kili.helpers import format_result, fragment_builder
+from kili.utils.pagination import api_throttle
+
+from ..graphql_client import GraphQLClient
```
PierreLeveau, line 13: you can move the `graphql_client.py` module into `src/kili/graphql/`

src/kili/graphql/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+class QueryOptions(NamedTuple):
+    """Options when calling GraphQLQuery from the SDK"""
+
+    first: Optional[int] = None
+    skip: int = 0
+    disable_tqdm: bool = True
+    as_generator: bool = False
+
+
+class GraphQLQuery(ABC):
```
PierreLeveau, line 25: 👍

src/kili/graphql/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+
+from ..graphql_client import GraphQLClient
+
+
+class QueryOptions(NamedTuple):
+    """Options when calling GraphQLQuery from the SDK"""
+
+    first: Optional[int] = None
+    skip: int = 0
+    disable_tqdm: bool = True
```
PierreLeveau, line 21: you can make `disable_tqdm` an `Optional[bool]`. `None` is accepted by tqdm, this disables tqdm when the code is run in non interactive mode (which should be our default btw)

src/kili/graphql/operations/project_user/queries.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+
+
+from typing import NamedTuple, Optional
+
+from kili.graphql import GraphQLQuery
+from kili.types import ProjectUser
+
+
+class ProjectUserWhere(NamedTuple):
```
PierreLeveau, line 12: alternatively, the conversion to a GraphQL "where" dict can be done in this class. This would be more self-contained.

src/kili/queries/project_user/__init__.py
```diff
@@ -43,7 +44,7 @@ def project_users(
         skip: int = 0,
         disable_tqdm: bool = False,
         as_generator: bool = False,
-    ) -> Union[List[dict], Generator[dict, None, None]]:
+    ):
```
PierreLeveau, line 47: why removing the return type ?

PierreLeveau (changes requested):

src/kili/cli/project/member/helpers.py
theodu: i tested, it in fact does not fail in the try when the project does not exist... I will change the error raised

src/kili/graphql/__init__.py
theodu: For inner queries, we should set disable_tqdm at True by default, and only activate it when called in a client function.
I set True by default here because we tend to increase the number of inner calls to the backend and like this we don't have to always import the QueryOptions class and disable tqdm

theodu: But that mean that for client functions, disable_tqdm should be set to None and to False ?

src/kili/graphql/operations/project_user/queries.py
theodu: Very good Idea! I didn't think about it

src/kili/queries/project_user/__init__.py
theodu: it is an error, I put in back with `Iterables[Dict]` as it is for other queries

src/kili/graphql/__init__.py
PierreLeveau: | The user is not interested by the fact that we do inner calls to Kili isn't it ?
we agree on that, this is where the `disable_tqdm` should be hard-coded to `True`.
I am not really annoyed by hard-coding it to true it each time we query 1 project. And I persist that in the other cases we should propagate what the user has decided (i.e. in the "main" query to Kili).

PierreLeveau: the reason I prefer explicit arguments in inner code is that if we change the default value one day for some reason, it will have side effects everywhere.

src/kili/graphql/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+    def execute_query_from_paginated_call(
+        self, query: str, where_payload: Dict, options: QueryOptions
+    ):
+        """
+        Builds a row generator from paginated calls.
+
+        Args:
+            skip: Number of assets to skip
+                (they are ordered by their date of creation, first to last).
+            first: Maximum number of assets to return.
```
PierreLeveau, line 125: docstring is not up-to-date

src/kili/graphql/operations/api_key/queries.py
```diff
@@ -0,0 +1,59 @@
...
+    Tuple to be passed to the APIKeyQuery to restrict the query
+    """
+
+    def __init__(
+        self,
+        api_key_id: Optional[str] = None,
+        user_id: Optional[str] = None,
+        api_key: Optional[str] = None,
+    ):
+        self.api_key_id = api_key_id
```
PierreLeveau, line 24: you probably should add the type, I am not sure it is propagated.

PierreLeveau: `self.api_key_id: Optional[str] = api_key_id`

src/kili/graphql/operations/project_user/queries.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    def __init__(
+        self,
+        project_id: str,
+        email: Optional[str] = None,
+        _id: Optional[str] = None,
+        organization_id: Optional[str] = None,
+    ):
+        self.project_id = project_id
```
PierreLeveau, line 24: same as above, you may add the types to the attribute init

PierreLeveau (changes requested):

src/kili/graphql/__init__.py
theodu: oh indeed!

src/kili/graphql/operations/api_key/queries.py
theodu: by hovering on the variable in the functions below it shows the type.
If it appears propagated, I prefer not to add it not to overload the code

PierreLeveau (approved):

PR #854, opened 2022-12-29
ci: add more info in create draft patch workflow

Jonas1312:


.github/workflows/release_draft_patch.yml
```diff
@@ -3,10 +3,10 @@ on:
   workflow_dispatch:
     inputs:
       releaseVersion:
-        description: "Latest minor release branch version"
+        description: "Latest minor release branch version (do not create a release branch for a patch!)"
```
theodu, line 6: I would rather put this "(do not create a release branch for a patch!)" on the release_branch workflow. If someone is very new, he would only see this message after having triggered the release_branch branch workflow on a patch.
I would also put in the description here the example of the release version to better visualize it: "X.XXX.0" (like you did in the release_draft_minor workflow)

theodu (approved):

theodu:
Très bien pour moi, je peux le merger ?

PR #857, opened 2022-12-29
feat: implement download_media for jsonContent assets

Jonas1312:


src/kili/queries/asset/helpers.py
```diff
@@ -73,30 +107,75 @@ def download_file(url: str, external_id: str, local_dir_path: Path) -> Path:
...
+
+    # put paths to files back into assets dicts
+    i = j = 0
+    while i < len(assets):
+        if assets[i]["content"] != "":
+            assets[i] = {**assets[i], "content": str(paths[j])}
+            j += 1
+        elif assets[i]["jsonContent"] != "":
+            if assets[i]["project"]["inputType"] == "VIDEO":
+                assets[i] = {**assets[i], "content": str(paths[j].parent)}  # root dir for frames
```
theodu, line 166: I think that we can't really put the folder name in the content field. it is not an informative field, but rather a technical field storing the information for the app on the asset media.
But I admit that it is ambiguous for local asset paths. We need to discuss this tomorrow at the daily.

src/kili/queries/asset/helpers.py
```diff
@@ -42,26 +43,59 @@ def get_file_extension_from_headers(url) -> Optional[str]:
...
+            return guess_extension(content_type)
+    return None
+
+
+def get_file_name_from_url(url) -> str:
+    """guess the filename of a file url"""
+    path = urlparse(url).path
+
+    # video frames
+    if "/api/label/v2/files" in path and "/frame/" in url:
```
theodu, line 57: This is true for frames hosted on Kili, i.e an asset imported from the UI. However, we can provide frames hosted elsewhere and they will have a different url format

theodu: I think that we have to stick with the externalId to infer the filename. We can provide an index as argument for frames to create a name like externalId/0001.png, externalId/0002.png etc
We do it in the UI for the Yolo export, you can have a look in the service label-export in the kili codebase.

src/kili/queries/asset/helpers.py
```diff
@@ -73,30 +107,75 @@ def download_file(url: str, external_id: str, local_dir_path: Path) -> Path:
...
-            raise NotImplementedError(
-                "The media of the asset is not stored in the content field. If your asset is video"
-                " imported from frames or a richText asset, the media is stored in the jsonContent"
-                " field. The download in local of such asset media is currently not supported."
+                " querying assets."
             )
 
 
+# pylint: disable=too-many-branches
 def download_asset_content(assets: List[dict], local_dir_path: Path) -> List[dict]:
```
theodu, line 120: the purpose of this function was to download content (see the function name). I would find it clearer to have two functions download_asset_content and download_asset_frames and to call the right one. (And see where to put rich text)

src/kili/queries/asset/helpers.py
```diff
@@ -73,30 +107,75 @@ def download_file(url: str, external_id: str, local_dir_path: Path) -> Path:
 
 def assert_required_fields_existence(asset: dict) -> None:
     """Get a list of an asset field from a generator of kili assets"""
-    required_fields = ["content", "externalId"]
+    required_fields = ["content", "externalId", "jsonContent", "project.inputType"]
```
theodu, line 110: jsonContent and project.inputTypes are not in the default fields when querying assets.
This change would be breaking: a client wanting to download classic assets (with just content) would see his script fail because here you refuse to download his assets because he didn't ask to query the field jsonContent.
I think that we can't force people whi want to query "classic" assets to query the jsonContent field

src/kili/queries/asset/helpers.py
theodu: I tested it, and the format used for the tree and names is:
Images/external_id1_0001.ext
Images/external_id1_0002.ext
...
Images/external_id2_0001.ext
Images/external_id2_0002.ext
...
...

So the tree is different that the choice that you take. We have to discuss tomorrow if it is smart to change that tree format in the assets queries compared to the export feature.

theodu (changes requested):
Some critic points. We should discuss my comments in a proper meeting to come up with a strategy

src/kili/queries/asset/helpers.py
Jonas1312: Better stay consistent with the UI for now

src/kili/queries/asset/helpers.py
Jonas1312: I added a warning if `jsonContent ` is not requested

PierreLeveau (changes requested):
`tests/queries/asset/__init__.py` can be removed

src/kili/queries/asset/helpers.py
```diff
@@ -49,133 +49,130 @@ def get_file_extension_from_headers(url) -> Optional[str]:
...
+                    "To download 'jsonContent' assets, please add 'project.inputType' to the fields"
+                    " when querying assets."
+                ) from exc
+
+
+def download_single_asset(asset: Dict, local_dir_path: Path) -> Dict[str, Any]:
+    """Download single asset on disk and modify asset attributes"""
+
+    if asset["content"] != "":
+        asset["content"] = download_file(asset["content"], asset["externalId"], local_dir_path)
```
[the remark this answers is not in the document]
theodu: In fact here it will download both, the frames and the original video. Don't know if we should do this

src/kili/queries/asset/helpers.py
```diff
@@ -49,133 +49,130 @@ def get_file_extension_from_headers(url) -> Optional[str]:
...
-def get_json_content_urls(json_url: str) -> Tuple[str]:
+def get_json_content_urls_video(json_url: str) -> Tuple[str]:
     """Get frame urls from a jsonContent url."""
     response = requests.get(json_url, timeout=20)
     response = response.json()
     urls = tuple(response.values())
     return urls
+
+
+def get_json_content_urls_image(json_url: str):
```
theodu, line 174: to be deleted

src/kili/queries/asset/helpers.py
```diff
@@ -49,133 +49,130 @@ def get_file_extension_from_headers(url) -> Optional[str]:
...
+            "It is recommended to add 'jsonContent' and 'project.inputType' to the 'fields'"
+            " argument in order to download all types of content."
+        )
+
+    else:
+        has_non_empty_jsoncontent = any(asset["jsonContent"] != "" for asset in assets)
+        if has_non_empty_jsoncontent:
+            try:
+                assets[0]["project"]["inputType"]
+            except KeyError as exc:
```
theodu, line 100: Since we force the user to give a project_id and that assets would all belong to the same project, I would query the project input type at the beginning of the download_assets function and not force him to ask it in the fields. Such a simple query should only add a very small amount of time compared to the media download

tests/e2e/test_query_assets_download_media.py
```diff
@@ -0,0 +1,437 @@
...
+        json_content_array = [
+            [os.path.join(tmp_dir, filename) for filename in sorted(os.listdir(tmp_dir))]
+        ]
+        kili.append_many_to_dataset(
+            project_id=project["id"],
+            content_array=None,
+            external_id_array=["vid 2"],
+            json_content_array=json_content_array,
+        )
+
```
theodu, line 161: add an asset with both content and json_content to simulate that case. (case where you import a video in mp4 format and ask the app to divide it in frames)

tests/e2e/test_query_assets_download_media.py
```diff
@@ -0,0 +1,437 @@
...
+        assets = kili.assets(
+            project_id=src_project_text["id"],
+            fields=["externalId", "content", "jsonContent", "project.inputType"],
+        )
+
+        assets = download_assets(assets, tmp_dir)
+
+        assert assets[0]["content"] == os.path.join(str(tmp_dir.resolve()), "text1")
+        assert Path(assets[0]["content"]).is_file()
+        assert md5_hash(Path(assets[0]["content"])) == "b11cc6e320a7bf4880c1d18714962f2d"
```
theodu, line 264: I don't get the purpose of hashing the path. Why do you use it?

theodu (changes requested):

src/kili/queries/asset/helpers.py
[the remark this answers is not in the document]
Jonas1312: you're right yes it will download both.

Maybe we will need the frames for the yolo, voc, exports no?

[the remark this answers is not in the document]
theodu: If we only download one among the two, it will be the frames and not the content. because the user annotate on the frames that we have split from the video.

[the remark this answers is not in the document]
PierreLeveau: I do not see a use case where the content would be more useful than the frames, so I am for outputting the frames.

theodu:
@Jonas1312 For me, a good strategy would be to:
1- fetch the type of the project
2- if it is frame or text and that the client did not query the jsonContent field, add it
3- download the media:
   - the json content if json content is not empty and the input type is TEXT
   - the json content if json content is not empty and the input type is FRAME
   - else: the content
  
4- if the jsonContent were not queried:
    - the you download the json content media, warn the user that you added it and that the info is inside
    - if you downloaded the content, remove the json content that were not asked by the client from the asset

happy to discuss it

Jonas1312:
e2e tests: https://github.com/kili-technology/kili-python-sdk/actions/runs/3845650819

Jonas1312:
e2e pass: https://github.com/kili-technology/kili-python-sdk/actions/runs/3846369044

src/kili/queries/asset/__init__.py
```diff
@@ -220,9 +220,20 @@ def assets(
             },
         }
 
-        post_call_process = get_post_assets_call_process(
-            download_media, local_media_dir, project_id
-        )
+        if download_media:
```
PierreLeveau, line 223: this starts to be a big method, so hiding the code added below in `get_post_assets_call_process` could be a good idea

PierreLeveau (changes requested):

Jonas1312:
e2e tests: https://github.com/kili-technology/kili-python-sdk/actions/runs/3847746903

PierreLeveau (approved):

tests/e2e/test_query_assets_download_media.py
```diff
@@ -0,0 +1,488 @@
...
+        width = 3800
+        height = 3800
+
+        img = Image.new(mode="RGB", size=(width, height))
+
+        # fill image with random data to avoid image compression
+        random_data = [random.randint(0, 255) for _ in range(width * height)]
+        img.putdata(list(zip(random_data, random_data, random_data)))  # type: ignore
+
+        img.save(temp, "PNG")
```
theodu, line 357: it could be coded within another PR but we should assert that the size of the created image is bigger than LARGE_IMAGE_THRESHOLD_SIZE to be sure that it is upload asynchronously by the backend

theodu (approved):

tests/e2e/test_query_assets_download_media.py
theodu: ah yes indeed!

PR #863, opened 2023-01-04
fix: duplicate external id in rich text notebook

theodu:


Jonas1312 (approved):

PierreLeveau:
alternatively `random.sample` can be used to draw integers without duplicate.

PierreLeveau (approved):

PR #866, opened 2023-01-05
fix: migrate labels query to the new architecture

theodu:


theodu:
- updated the way we do pagination. We know count the number of objects to query and then do paginated calls until to have this number instead of querying objects until to have nothing.
- solve an bug showing the wrong number of fetched objects when querying with a skip option
- migrate the labels query
- write tests for the new queries architecture

src/kili/graphql/__init__.py
```diff
@@ -125,25 +123,24 @@ def execute_query_from_paginated_call(
...
                     skip = count_rows_retrieved + options.skip
-                    payload.update({"skip": skip, "first": batch_size})
+                    payload = {"where": where.graphql_payload, "skip": skip, "first": batch_size}
                     rows = api_throttle(self.client.execute)(query, payload)
                     rows = format_result("data", rows)
 
-                    if rows is None or len(rows) == 0:
-                        break
+                    # if rows is None or len(rows) == 0:
+                    #     break
```
Jonas1312, line 143: remove comments?

Jonas1312 (approved):

src/kili/graphql/__init__.py
theodu: I will add it back. It is useless but could still act as a safe line to keep

PierreLeveau (approved):

PR #869, opened 2023-01-06
fix: fix query in update_properties_in_project

Jonas1312:


theodu (changes requested):
No this is not the problem. I will correct it in another PR

Jonas1312:
it fixes this bug:
```python
kili.update_properties_in_project(
        project_id="",
        metadata_types={
            "accuracy": "number",
        },
    )
```

```shell
kili.exceptions.NonExistingFieldError: Cannot query field metadataTypes on object Project. Admissible fields are:
```

PR #874, opened 2023-01-09
chore: migrate queries into graphql folder: Issues, Locks

theodu:


src/kili/queries/issue/__init__.py
```diff
@@ -1,14 +1,13 @@
...
 
-from ...helpers import format_result, fragment_builder
-from ...types import Issue as IssueType
-from ...utils.pagination import row_generator_from_paginated_calls
-from .queries import GQL_ISSUES_COUNT, gql_issues
+from kili.graphql import QueryOptions
+from kili.graphql.operations.issue.queries import IssueQuery, IssueWhere
 
 
 @dataclass
```
PierreLeveau, line 13: remove `@dataclass` ?

PierreLeveau (approved):

src/kili/queries/lock/__init__.py
```diff
@@ -29,6 +28,10 @@ def __init__(self, auth):
 
     # pylint: disable=dangerous-default-value
     @typechecked
+    @deprecate(
+        "locks method is for internal use only and will be removed from 01/02/2013",
```
PierreLeveau, line 32: I don't think there would be a use case for locks. We can skip the deprecation period.

theodu: We want to keep it internally but I deprecated it for the sake of not introducing a breaking change without a deprecation period. 
I checked in Apollo, it has not been called in the last 3 months.
If you feel safe not doing a deprecation period, I follow you.

PierreLeveau (changes requested):

src/kili/graphql/operations/issue/queries.py
```diff
@@ -0,0 +1,52 @@
...
+from kili.graphql import BaseQueryWhere, GraphQLQuery
+from kili.types import Issue
+
+
+class IssueWhere(BaseQueryWhere):
+    """
+    Tuple to be passed to the IssueQuery to restrict query
+    """
+
+    # pylint: disable=too-many-arguments,too-many-locals,too-many-instance-attributes
```
Jonas1312, line 16: we can remove that flag

@PierreLeveau we should consider removing those pylint flags (too-many-...). We ignore them anyway, so better disabling them from linting settings maybe

src/kili/graphql/operations/lock/queries.py
```diff
@@ -0,0 +1,52 @@
...
+from kili.graphql import BaseQueryWhere, GraphQLQuery
+from kili.types import Lock
+
+
+class LockWhere(BaseQueryWhere):
+    """
+    Tuple to be passed to the LockQuery to restrict query
+    """
+
+    # pylint: disable=too-many-arguments,too-many-locals,too-many-instance-attributes
```
Jonas1312, line 16: same

src/kili/graphql/operations/issue/queries.py
PierreLeveau: i would say it depends... For internal code, especially the core, I think it is interesting to keep them as it underlines good practices. For external code, it is almost a design choice to pass flat argument lists so we should silence them.

theodu: We already discussed this. I think that Pierre was in favor of keeping them so it applies to internal functions and we keep it for user-facing functions

PierreLeveau (approved):

Jonas1312 (approved):

PR #880, opened 2023-01-09
feat(DATA-892): upload directory as plugin_src

pmKili:


src/kili/services/plugins/upload.py
```diff
@@ -102,20 +157,42 @@ def _retrieve_upload_url(self, is_updating_plugin: bool) -> str:
...
+        """
+        Create a zip file from python file and requirements.txt
+        (if user has defined a path to a requirements.txt)
+        """
+        zip_path = tmp_directory / "archive.zip"
+
+        with ZipFile(zip_path, "w") as archive:
+            archive.write(script_path, "main.py")
+            if requirements_path:
+                archive.write(requirements_path)
```
gheorghetutunaru, line 175: Shouldn't this also have a second argument "requirements.txt" as does the call for "main.py" ?

src/kili/services/plugins/upload.py
```diff
@@ -18,78 +19,132 @@
...
 NUMBER_TRIES_RUNNER_STATUS = 20
 
 
-def check_file_is_py(path: Path, verbose: bool = True) -> bool:
+def check_file_mime_type(
+    path: Path, compatible_mime_extensions: List[str], verbose: bool = True
+) -> bool:
     """
-    Returns true if the mime type of the file corresponds to a python file
+    Returns true if the mime type of the file corresponds to a compatible_mime_extensions
```
gheorghetutunaru, line 33:
```suggestion
    Returns true if the mime type of the file corresponds to one of compatible_mime_extensions
```

src/kili/services/plugins/upload.py
pmKili: I did it for `main.py` because if the user choose a path to a python file, the file could be named differently than `main.py`. While here `requirements.txt` could only be named `requirements.txt`

src/kili/services/plugins/upload.py
pmKili: done!

gheorghetutunaru (approved):

src/kili/mutations/plugins/__init__.py
```diff
@@ -123,14 +123,16 @@ def delete_plugin(
     @typechecked
     def update_plugin(
         self,
-        file_path: str,
+        plugin_path: str,
```
p-desaintchamas, line 126: Seen with you, good to have a proper error message on our end so older sdk version know they have to upgrade kili

pmKili: done!

PR #882, opened 2023-01-10
docs: add tool to convert notebook to markdown

Jonas1312:
new pre commit hooks:
- one for checking if notebook is formatted with black (only applies to notebooks in `/docs/sdk/tutorials`)
- another one that triggers during commit when notebook or markdown files are modified in `/docs/sdk/tutorials`. The hook will check if the notebook is up-to-date with the markdown generated file. It will also check if the markdown is present in `mkdocs.yml`

One command `python -m docs.utils convert tuto.ipynb` to convert `tuto.ipynb` notebook to markdown `tuto.md`.

Images are included in the markdown directly as base64 blobs.

### DIFF
--- a/docs/utils.py
+++ b/docs/utils.py
@@ -196,6 +196,9 @@ def notebook_tutorials_commit_hook(modified_files: Sequence[Path]):
         # check only notebooks modified in docs/sdk/tutorials
         if "docs/sdk/tutorials" not in str(group[0].parent):
             continue
+        # skip single markdown files, probably hand written tutorials without notebooks
+        if len(group) == 1 and str(group[0]).endswith(".md"):
+            continue
         if len(group) != 2:
             raise ValueError(
                 f"Expected two files (.md and .ipynb) in staging for '{filename}', got {group}."
--- a/docs/utils.py
+++ b/docs/utils.py
@@ -2,6 +2,7 @@
 
 import base64
 import re
+import shutil
 from binascii import a2b_base64
 from itertools import groupby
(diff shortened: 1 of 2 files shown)


PierreLeveau (approved):

docs/utils.py
```diff
@@ -0,0 +1,235 @@
...
+
+        ipynb_filepath = ipynb_filepath.resolve()
+        md_filepath = md_filepath.resolve()
+
+        if not is_markdown_up_to_date(
+            ipynb_filepath=ipynb_filepath,
+            md_filepath=md_filepath,
+            remove_cell_tags=DEFAULT_REMOVE_CELL_TAGS,
+        ):
+            raise OutdatedMarkdownError(f"{ipynb_filepath} is not up to date with {md_filepath}.")
```
theodu, line 218: why do you raise an error instead of replacing the markdown with the one that you generated to test the equality?

theodu: like this, we wouldn't need to call notebook_to_markdown_cmd manually.

theodu (commented):
Super good! just a question on why we do not automatically regenerate the markdown when the change the jupyter file

docs/utils.py
theodu: I understand. But then we could not preview the markdown on Mkdocs to assert that everything is good. If we are currently not sure of the quality of generation, I am ok to begin to generate markdown manually with the command and then do it manually if we are more confident

theodu (approved):

PR #883, opened 2023-01-10
ci: increase timeout for unit tests

Jonas1312:
python 3.11 tests take more than 4 minutes sometimes: https://github.com/kili-technology/kili-python-sdk/actions/runs/3872411293


theodu (commented):
I validate but we should keep an eye on the time taken for the tests. Not to automatically increase the time when the tests timeout and sometime try to optimize them.
Here I think that the main cause is that we add more and more tests so it is fine!

PR #891, opened 2023-01-12
test: fix windows unit testing

Jonas1312:
Windows specificities:
- `requests.get()` needs `os.environ` keys, be careful when mocking
- always create the directory before moving/copying/creating a file into it
- use `os.sep` to get the windows `\\` or unix `/` separators
- `csv.writer` requires a file object with `newline=""` : https://stackoverflow.com/questions/3191528/csv-in-python-adding-an-extra-carriage-return-on-windows
- sometimes `print()` fails to encode characters when windows cmd is set to CP-1252 encoding

TODO:
- [x] make all unit tests pass
- [x] trigger e2e tests before merging: https://github.com/kili-technology/kili-python-sdk/actions/runs/3904449606

### DIFF
--- a/src/kili/services/export/format/coco/__init__.py
+++ b/src/kili/services/export/format/coco/__init__.py
@@ -227,7 +227,7 @@ def _get_coco_images_and_annotations(
         coco_image = _CocoImage(
             id=asset_i,
             license=0,
-            file_name=str("data" / Path(file_name.parts[-1])),
+            file_name=str("data" + "/" + file_name.parts[-1]),
             height=height,
             width=width,
             date_captured=None,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,12 +22,18 @@ jobs:
         run: |
           pylint src/kili
   unittest:
-    runs-on: ubuntu-latest
     timeout-minutes: 6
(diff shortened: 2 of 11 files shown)


src/kili/services/export/format/coco/__init__.py
```diff
@@ -227,7 +227,7 @@ def _get_coco_images_and_annotations(
         coco_image = _CocoImage(
             id=asset_i,
             license=0,
-            file_name=str("data" / Path(file_name.parts[-1])),
+            file_name=str("data" + "/" + file_name.parts[-1]),
```
PierreLeveau, line 230:
```suggestion
            file_name=str(PosixPath("data") / file_name.parts[-1]),
```
should be working

PierreLeveau (changes requested):

PierreLeveau (approved):

theodu (approved):

src/kili/services/export/format/coco/__init__.py
PierreLeveau: indeed this is `PurePosixPath` . I let you choose the solution. In any case, you should add a comment to says that the posix format is imposed by the COCO format.

PR #895, opened 2023-01-13
fix: Error message is sent when trying to upload plugin wit…

hguerni:


src/kili/services/plugins/upload.py
```diff
@@ -62,6 +64,26 @@ def check_file_is_txt(path: Path, verbose: bool = True) -> bool:
...
+    """
+    spec = importlib.util.spec_from_file_location(name, path)
+    if not spec:
+        return False
+    plugin = importlib.util.module_from_spec(spec)
+    sys.modules[name] = plugin
+    loader = spec.loader
+    if not loader:
+        return False
+    loader.exec_module(plugin)
```
p-desaintchamas, line 79: try leaving some empty lines to make your code more readable

p-desaintchamas (changes requested):

src/kili/services/plugins/upload.py
hguerni: done !

hguerni:
I have been trying to rebase from master to fix conflict, but every time I fix it it seems to revert to previous state.

PR #899, opened 2023-01-16
docs: add plugins tutorials to SDK doc

Jonas1312:


docs/sdk/tutorials/plugins_development.md
```diff
@@ -0,0 +1,284 @@
...
+```python
+from kili.exceptions import GraphQLError
+
+try:
+    kili.upload_plugin(path_to_plugin, plugin_name)
+except GraphQLError as error:
+    print(str(error))
+```
+
+    Hint: A plugin with this name already exist, if you want to override it you can use the command kili.update_plugin(plugin_path="plugin.py", plugin_name="My first kili plugin")
```
theodu, line 202: error here, we want to get rid of this isn't it ?

docs/sdk/tutorials/plugins_development.md
```diff
@@ -0,0 +1,284 @@
+<!-- FILE AUTO GENERATED BY docs/utils.py DO NOT EDIT DIRECTLY -->
+<a href="https://colab.research.google.com/github/kili-technology/kili-python-sdk/blob/master/recipes/tutorials/plugins_development.ipynb" target="_parent"><img src="https://colab.research.google.com/assets/colab-badge.svg" alt="Open In Colab"/></a>
```
theodu, line 2: 'Notebook introuvable"

theodu (changes requested):
Other comment: I don't get why we add 2 tutorials about plugins that basically contain almost the same information and have the same sections

docs/sdk/tutorials/plugins_development.md
Jonas1312: I can silence the error, not sure what's best @xczuba

Jonas1312:
> Other comment: I don't get why we add 2 tutorials about plugins that basically contain almost the same information and have the same sections

I totally agree, they are really similar...

theodu:
> > Other comment: I don't get why we add 2 tutorials about plugins that basically contain almost the same information and have the same sections
> 
> I totally agree, they are really similar...

This is disturbing for the client because he has to jump back and forth between the two tutorials to compare the information. Can it be seen with @xczuba to merge it into one?

docs/sdk/tutorials/plugins_development.md
p-desaintchamas: @Jonas1312 @theodu If those are generated, is it really necessary to commit them ? ❓

Jonas1312: not necessary per se, but it's safer that way since the notebook to markdown conversion is still new and probably not bug free.

It's better to force the PR author to include the markdown and double-check the render in mkdocs

PierreLeveau (approved):

theodu (approved):

PR #900, opened 2023-01-16
feat: add webhook creation & update

p-desaintchamas:


src/kili/mutations/plugins/__init__.py
```diff
@@ -59,6 +60,64 @@ def upload_plugin(
...
+    @typechecked
+    def create_webhook(
+        self,
+        webhook_url: str,
+        plugin_name: str,
+        header: Optional[str] = None,
+        verbose: bool = True,
+    ):
+        # pylint: disable=line-too-long
+        """Create a webhook linked to kili's event to receive custom events.
```
pmKili, line 72:
```suggestion
        """Create a webhook linked to kili's events.
```

src/kili/mutations/plugins/__init__.py
```diff
@@ -59,6 +60,64 @@ def upload_plugin(
...
+    @typechecked
+    def update_webhook(
+        self,
+        webhook_url: str,
+        plugin_name: str,
+        header: Optional[str] = None,
+        verbose: bool = True,
+    ):
+        # pylint: disable=line-too-long
+        """Create a webhook linked to kili's event to receive custom events.
```
pmKili, line 101:
```suggestion
        """Update in the webhook the url where kili will do HTTP requests
```

src/kili/mutations/plugins/__init__.py
```diff
@@ -59,6 +60,64 @@ def upload_plugin(
...
+
+        Args:
+            webhook_url: Url base receiving push events from on_submit / on_review events on Kili
```
pmKili, line 104: I think it would be better to use `new_webhook_url` so that users understand it's the new URL that they're supposed to provide

pmKili (approved):

PR #903, opened 2023-01-16
feat: AAU, I can import a folder as a plugin

pmKili:


florianlega (approved):

src/kili/services/plugins/upload.py
```diff
@@ -279,6 +293,9 @@ def _check_plugin_runner_status(self, update=False):
             if status == "ACTIVE":
                 break
 
+            if status == "FAILED":
```
gheorghetutunaru, line 296: It would be a bit cleaner to just add ` or status == "FAILED"` two lines above

src/kili/services/plugins/upload.py
```diff
@@ -214,18 +218,26 @@ def _retrieve_upload_url(self, is_updating_plugin: bool) -> str:
...
+            if len(file_paths) == 0:
+                archive.write(file_paths[0], "main.py")
+
+            else:
+                for path in file_paths:
+                    archive.write(path, path.relative_to(self.plugin_path))
+
             if requirements_path:
-                archive.write(requirements_path, "requirements.txt")
+                archive.write(requirements_path, requirements_path.relative_to(self.plugin_path))
```
gheorghetutunaru, line 240: Why does it need to be relative and not just at the root of the zip ?

src/kili/services/plugins/upload.py
pmKili: indeed 😅

src/kili/services/plugins/upload.py
pmKili: It's because if I have a folder like that:
- plugin-folder
   - main.py
   - requirements.txt
   - sub-folder
       - helpers.py
       
The value of `requirements_path` would be `plugin-folder/requirements.txt` while `requirements_path.relative_to(self.plugin_path) = 'requirements.txt'`

However, this line change is useless. I did `archive.write(requirements_path, "requirements.txt")` today on another PR to fix the upload of requirements.txt while I did `archive.write(requirements_path, requirements_path.relative_to(self.plugin_path))` yesterday. Both way are doing the same thing

p-desaintchamas (approved):

PR #910, opened 2023-01-17
feat: add video export

Jonas1312:


tests/e2e/test_copy_project.py
```diff
@@ -395,8 +396,7 @@ def test_copy_project_e2e(kili, src_project):
         for asset_src, asset_new in zip(assets_src, assets_new):
             assert asset_src["externalId"] == asset_new["externalId"]
 
-            # Path("https://...").is_file() crashes on windows with python 3.7
-            if not asset_src["content"].startswith("http") and Path(asset_src["content"]).is_file():
+            if not asset_src["content"].startswith("http") and os.path.isfile(asset_src["content"]):
```
PierreLeveau, line 399: we should leave the comment  to explain why we do not use `Path.is_file()`

PierreLeveau (changes requested):

tests/e2e/test_copy_project.py
Jonas1312: os.path.is_file() works well on windows, is there a reason we shouldn't use it?

src/kili/services/export/format/kili/__init__.py
```diff
@@ -43,12 +54,49 @@ def _save_assets_export(self, assets: List[Asset], output_filename: Path):
...
         self.make_archive(self.export_root_folder, output_filename)
 
         self.logger.warning(output_filename)
 
+    def _clean_filepaths(self, assets: List[Asset]):
+        """
+        Remove TemporaryDirectory() prefix from filepaths in "jsonContent" and "content" fields.
+        """
+        for asset in assets:
+            if asset["content"] != "" and os.path.isfile(asset["content"]):
```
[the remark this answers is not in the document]
Jonas1312: os.path.isfile() does the same thing? but has the advantage to work on windows unlike Path.is_file

tests/e2e/test_copy_project.py
PierreLeveau: I agree, but someone might forget that there is a good reason not to use Pathlib, and will make the change the ther way around.

PierreLeveau (approved):

PR #912, opened 2023-01-18
feat: cannot upload plugin without pluginHandler class, error message is sent

hguerni:


src/kili/services/plugins/upload.py
```diff
@@ -156,6 +182,9 @@ def _retrieve_script(self) -> Path:
         if not check_file_is_py(file_path, self.verbose):
             raise ValueError("Wrong file format.")
 
+        if not check_file_contain_handler(self.plugin_name, str(file_path)):
+            raise ValueError("Do not contain PluginHandler Class. ")
+
```
pmKili, line 187: Can you do that in `_parse_script` instead?

hguerni: yes will do

src/kili/services/plugins/upload.py
```diff
@@ -181,16 +212,18 @@ def _retrieve_requirements(self) -> Union[Path, None]:
         return file_path
 
     @staticmethod
-    def _parse_script(script_path: Path):
+    def _parse_script(script_path: Path, name: str):
         """
         Method to detect indentation errors in the script
         """
+        if not check_file_contain_handler(name, str(script_path)):
```
pmKili, line 219: check_file_contains_handler + you should enter only if the file is main.py

src/kili/services/plugins/upload.py
```diff
@@ -181,16 +212,18 @@ def _retrieve_requirements(self) -> Union[Path, None]:
...
+            raise ValueError("Do not contain PluginHandler Class. ")
```
pmKili, line 220:
```suggestion
            raise ValueError("Your main.py file does not contain a PluginHandler ")
```

src/kili/services/plugins/upload.py
```diff
@@ -186,11 +212,16 @@ def _retrieve_requirements(self) -> Union[Path, None]:
...
 
     @staticmethod
-    def _parse_script(file_path: Path):
+    def _parse_script(script_path: Path, name: str):
         """
-        Method to detect indentation errors in the script
+        Method to detect indentation and class errors in the script
         """
-        with file_path.open("r", encoding="utf-8") as file:
+        if str(script_path) == "main.py" and not check_file_contains_handler(
```
gheorghetutunaru, line 219: If this is a Path, couldn't it contain more elements ? I mean for example `path/to/folder/main.py` so your check on main would not work ?

pmKili: Yes, that's right. @hguerni did you test your feature before pushing your branch?

hguerni: Yes I did but only with file in root, now done using path.name

gheorghetutunaru (approved):

p-desaintchamas (approved):

PR #918, opened 2023-01-19
docs: delete notebooks in recipes/tutorials

Jonas1312:
Seems like it's better to keep all notebooks in recipes/ since we gave clients direct links to them.

This PR removes the recipes/tutorials folder and changes the pre-commit hook accordingly.

However, if we merge this, we need to update the documentation since the google colab links in the documentation point to the notebooks in recipes/tutorials


PierreLeveau (approved):

theodu (approved):
I approve to move on but as said in the daily, we should discuss with Damien about the issue

PR #919, opened 2023-01-19
ci: add ci for deploying doc

Jonas1312:
new deploy_doc.yml workflow:
- runs on workflow_dispatch to be able to trigger it manually
- runs on workflow_call to allow it being called by publish.yml
- I am not sure that the deploy_doc.yml will checkout the repository on the release branch. So I added a step to get the latest release version, and I checkout the proper release branch

in publish.yml
- removed the doc deployment related code
- calls the deploy_doc.yml workflow when the `publish` job is finished, whether it succeeded or not

### DIFF
--- a/.github/workflows/deploy_doc.yml
+++ b/.github/workflows/deploy_doc.yml
@@ -23,7 +23,7 @@ jobs:
           echo "REF=master" >> $GITHUB_ENV
 
       - name: set checkout to latest release branch
-        if: ${{ (github.event_name == 'workflow_dispatch' && github.event.inputs.branch == 'latest minor release branch') || github.event_name == 'workflow_call' }}
+        if: ${{ github.event_name == 'workflow_call' || (github.event_name == 'workflow_dispatch' && github.event.inputs.branch == 'latest minor release branch') }}
         run: |
           last_version_tag=$(curl --silent "https://api.github.com/repos/kili-technology/kili-python-sdk/releases/latest" | jq -r .tag_name)
           echo $last_version_tag
--- a/.github/workflows/deploy_doc.yml
+++ b/.github/workflows/deploy_doc.yml
@@ -1,0 +1,52 @@
+name: Deploy SDK doc
+on:
+  workflow_dispatch:
+
+jobs:
+  deploy-documentation:
+    name: Deploy technical documentation
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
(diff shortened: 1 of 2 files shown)


theodu (changes requested):
we should call this workflow inside the publish workflow to factories the code in case we make changes to the workflow

PierreLeveau (approved):

Jonas1312:
@PierreLeveau @theodu when we trigger the deploy_doc.yml, we expect it to release the doc on the latest release branch right? Not master?

.github/workflows/deploy_doc.yml
```diff
@@ -0,0 +1,64 @@
...
+jobs:
+  deploy-documentation:
+    name: Deploy technical documentation
+    runs-on: ubuntu-latest
+    steps:
+      - name: get last release number
+        run: |
+          last_version_tag=$(curl --silent "https://api.github.com/repos/kili-technology/kili-python-sdk/releases/latest" | jq -r .tag_name)
+          echo $last_version_tag
+          echo "LAST_VERSION=$last_version_tag" >> $GITHUB_ENV
```
theodu, line 15: ah and also I haven't seen this: Here it deploys the doc from the latest release tag.
This is already done automatically in the publish script.
For me, adding a button in Github would only be useful if we can deploy the doc for another branch. For example the use case that Paul just had: I realize that I miss a part of the doc that is on master but not on the last release, then I want to release the doc from master

theodu (changes requested):

.github/workflows/deploy_doc.yml
theodu: Or maybe a miss something? Why would it be useful to have a button in Github to publish the doc from the last release if it is already automatically released ?

theodu: Ah I just saw that you raised the issue in your previous comment ;)

Jonas1312: so the workflow has:
- a workflow_call trigger for when it's called from publish.yml. The doc will be deployed from the release branch (same behavior as before)
- a workflow_dispatch to trigger the workflow from the UI. With the option to choose between checkout to master or checkout to latest release branch (default to master)

theodu (approved):
Super clean!!

.github/workflows/deploy_doc.yml
PierreLeveau: 👌

PierreLeveau (approved):

PR #939, opened 2023-01-23
ci: add pre-commit hook for markdown formatting

Jonas1312:
I tried markdown formatting with mdformat: https://mdformat.readthedocs.io/en/stable/

Too many incompatibilities between markdown standards (commonmark, github markdown, mkdocs markdown). mdformat often breaks mkdocs related commands (!!! note, !!! warning). 

We can't disable formatting on some lines with mdformat, so I had to exclude some markdown files to not break the SDK doc.

It also doesn't work well with nbconvert. So we can't use it to format the notebook and markdown tutorials.

I also tried some markdown linters, but they are often too pedantic.

I'm not sure if we should merge this since markdown formatting is not exact science...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,28 +1,40 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.1.0
+    rev: v4.4.0
     hooks:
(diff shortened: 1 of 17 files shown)


PierreLeveau:
I agree that's not necessary.

PR #942, opened 2023-01-23
fix: E2E tests with label validation

theodu:


theodu:
- https://gitlab.com/kili-technology/kili/-/merge_requests/9113 should be merged before for E2E tests to pass
- change the workflow recipe because interface was not in line with labels pushed

PierreLeveau (approved):

Jonas1312 (approved):

PR #947, opened 2023-01-24
fix: change css font size on the SDK doc website

theodu:


Jonas1312:
looks better IMO.

I wonder if we shouldn't also  reduce the TOCs font size (currently 16.8px) ?  Maybe 16px or 15px?

It's defined in `.md-nav`

theodu:
Yes it looks better now!
When I inspect, I see the nav font-size at `14px`, and it seems good. Strange that you see it at `16.8px`, can you send a screenshot ?

theodu:
oh, I don't have this fonts tab, it is useful!
Personally, I find the size of the nav good that way and I wouldn't change it. What do you think @PierreLeveau  ?

PierreLeveau (approved):

PierreLeveau:
LGTM

PR #953, opened 2023-01-27
chore: make MutationsOrganization internal use only

Jonas1312:


src/kili/mutations/organization/__init__.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, auth):
     @typechecked
     def create_organization(self, name: str, address: str, zip_code: str, city: str, country: str):
         """Create an organization.
-        WARNING: This resolver is for internal use by Kili Technology only.
+        WARNING: This method is for internal use by Kili Technology only.
```
PierreLeveau, line 28:
```suggestion
        WARNING: This method is for internal use only.
```

src/kili/mutations/organization/__init__.py
```diff
@@ -64,7 +64,7 @@ def update_properties_in_organization(
         license: Optional[dict] = None,
     ):  # pylint: disable=redefined-builtin
         """Modify an organization.
-        WARNING: This resolver is for internal use by Kili Technology only.
+        WARNING: This method is for internal use by Kili Technology only.
```
PierreLeveau, line 67:
```suggestion
        WARNING: This method is for internal use only.
```

PierreLeveau (changes requested):

PR #956, opened 2023-01-27
chore: 🤖 fetch upload url after successfully parsing the script

gheorghetutunaru:


src/kili/services/plugins/upload.py
```diff
@@ -275,15 +275,15 @@ def _upload_script(self, is_updating_plugin: bool):
...
-        upload_url = self._retrieve_upload_url(is_updating_plugin)
-
         file_paths = self._retrieve_plugin_src()
 
         for path in file_paths:
             self._parse_script(path, self.plugin_name)
 
         requirements = self._retrieve_requirements()
 
+        upload_url = self._retrieve_upload_url(is_updating_plugin)
```
gheorghetutunaru, line 285: Before if the parsing failed, we still created the Plugin entity even if we don't upload it

pmKili (approved):

p-desaintchamas (approved):

PR #960, opened 2023-01-30
fix: disable copy project for assets on remote storage

Jonas1312:
stop copy if project has a remote storage

### DIFF
--- a/src/kili/services/copy_project/__init__.py
+++ b/src/kili/services/copy_project/__init__.py
@@ -24,7 +24,6 @@ class ProjectCopier:  # pylint: disable=too-few-public-methods
         "inputType",
         "description",
         "id",
-        "dataConnections.id",
         "dataConnections.dataIntegrationId",
     ]
     FIELDS_JSON_INTERFACE = ["jsonInterface"]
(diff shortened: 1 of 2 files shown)


src/kili/services/copy_project/__init__.py
```diff
@@ -19,7 +19,14 @@ class ProjectCopier:  # pylint: disable=too-few-public-methods
...
     """
 
-    FIELDS_PROJECT = ["title", "inputType", "description", "id"]
+    FIELDS_PROJECT = [
+        "title",
+        "inputType",
+        "description",
+        "id",
+        "dataConnections.id",
+        "dataConnections.dataIntegrationId",
```
PierreLeveau, line 28: not necessary ?

PierreLeveau (approved):

PierreLeveau (approved):

theodu (commented):
This is indeed the best way to do it. 
But we could think of a client who has data both from remote storage and on a Kili Bucket. Copying all assets that are on the Kili bucket would allow him to just add his connection afterward in order to copy the project with all his assets, without manipulating the assets on his remote storage.  
I don't know if this situation would often arise. I am okay with this implementation because it is a fast fix and might handle most of the cases but we should also validate it with @hugo-mailfait-kili and @benjamin

PR #963, opened 2023-01-31
chore: 🤖 fix notebooks test for example and ignore development

gheorghetutunaru:


tests/test_notebooks.py
```diff
@@ -5,6 +5,10 @@
 import pytest
 from nbconvert.preprocessors.execute import ExecutePreprocessor
 
+IGNORED_NOTEBOOKS = [
+    "recipes/plugins_development.ipynb",
+]
```
gheorghetutunaru, line 10: Needed to not throw error that notebook is not tested. I could also leave the notebook in the list of `process_notebook` and then check if the name is not in the ignored list. Tell me what you prefer

Jonas1312: You can do it this way. Or add the IGNORED_NOTEBOOKS in check_notebook_tested instead: https://github.com/kili-technology/kili-python-sdk/blob/a70cb0f6cfd75c9f5aa2dc4ad83b98e0c7020d90/docs/utils.py#L245  So that the pre-commit hook skips the check_notebook_tested step for this tutorial

Jonas1312 (approved):

florianlega (approved):

Jonas1312 (changes requested):
notebook has a dead link: https://raw.githubusercontent.com/kili-technology/kili-python-sdk/master/recipes/datasets/plugins/plugin.py

Jonas1312 (approved):

PR #964, opened 2023-01-31
feat: add consensus plugin as example

hguerni:


recipes/datasets/plugins/plugin_consensus.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+
+    def aggregate_object_detection_label(self, object_job_key, labels_array):
+
+        json_response_array = {}
+
+        for key in object_job_key:
+            l_key = []
+            for _, label in enumerate(labels_array):
+                for annotation in label["jsonResponse"][key]["annotations"]:
```
gheorghetutunaru, line 17: I think this will fail if there is a label that does not have an annotation for each job (for example if a job is not required), check that `key` is in `label["jsonResponse"]`

recipes/datasets/plugins/plugin_consensus.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Custom plugin instance
+    """
+
+    def aggregate_object_detection_label(self, object_job_key, labels_array):
+
+        json_response_array = {}
+
+        for key in object_job_key:
+            l_key = []
```
gheorghetutunaru, line 15: `l_key` is not really clear, maybe `job_annotations` is a better name ?

recipes/datasets/plugins/plugin_consensus.py
```diff
@@ -0,0 +1,66 @@
...
+            for _, label in enumerate(labels_array):
```
gheorghetutunaru, line 16: why not `for label in labels_array` if you don't need the index anyway  ?

recipes/datasets/plugins/plugin_consensus.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class PluginHandler(PluginCore):
+    """
+    Custom plugin instance
+    """
+
+    def aggregate_object_detection_label(self, object_job_key, labels_array):
+
+        json_response_array = {}
```
gheorghetutunaru, line 12: confusing name, it's not an array

recipes/datasets/plugins/plugin_consensus.py
```diff
@@ -0,0 +1,66 @@
...
+            asset_id=asset_id,
+            type_in=["DEFAULT"],
+            fields=[
+                "author.email",
+                "author.id",
+                "id",
+                "jsonResponse",
+                "labelType",
+                "secondsToLabel",
+                "isLatestLabelForUser",
```
gheorghetutunaru, line 51: If you only need `isLatestLabelForUser` and `jsonResponse` in the plugin, there is no need to fetch extra fields (you can probably keep `id`, but fetching info on `author` makes an unnecessary join in the DB

recipes/datasets/plugins/plugin_consensus.py
```diff
@@ -0,0 +1,63 @@
...
+            type_in=["DEFAULT"],
+            fields=[
+                "id",
+                "jsonResponse",
+                "isLatestLabelForUser",
+            ],
+        )
+
+        labels_array_last = list({})
+        for label_iteration in labels_array:
```
p-desaintchamas, line 53: https://www.programiz.com/python-programming/methods/built-in/filter

latest_labels = filter(x: x["isLatestLabelForUser"] is True, labels_array)

recipes/datasets/plugins/plugin_consensus.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+
+    def aggregate_object_detection_label(self, object_job_key, labels_array):
+
+        json_response = {}
+
+        for key in object_job_key:
+            job_annotation = []
+            for label in labels_array:
+                if label["jsonResponse"][key]:
```
gheorghetutunaru, line 17: in python it's not the same as in js: 
image
image

hguerni: it's weird I don't get any errors

hguerni: Maybe it's a dictionnary thing idk

p-desaintchamas (approved):

florianlega (approved):

PR #965, opened 2023-01-31
chore: add tests for plugins

pmKili:


tests/e2e/plugin_workflow.ipynb
```diff
@@ -0,0 +1,309 @@
...
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/paulm/Documents/Git/GitHub_Kili/kili-playground/src/kili/authentication.py:128: UserWarning: \n",
+      "Your api key will be deprecated on 2023-02-16.\n",
+      "You should generate a new one on My account > API KEY.\n",
+      "  warnings.warn(message, UserWarning)\n"
+     ]
+    }
+   ],
```
gheorghetutunaru, line 40: Clear outputs

tests/e2e/plugin_workflow.ipynb
```diff
@@ -0,0 +1,309 @@
...
+   "source": [
+    "kili = Kili(\n",
+    "    api_endpoint=\"https://staging.cloud.kili-technology.com/api/label/v2/graphql\",\n",
+    "    api_key=os.getenv(\"KILI_ADMIN_API_KEY_STAGING\"),\n",
+    ")"
```
gheorghetutunaru, line 45: I think you can just use `kili = Kili() `and in the e2e tests there are the proper variables for api_key and endpoint

tests/services/plugins/plugin_folder/sub_folder/helpers.py
```diff
@@ -0,0 +1,2 @@
+def custom_function():
+    pass
```
gheorghetutunaru, line 2: Maybe it would be good to also call this function in the `main.py` to test sub-folders in the e2e test (that the build doesn't fail)

tests/e2e/plugin_workflow.ipynb
```diff
@@ -0,0 +1,309 @@
+{
```
gheorghetutunaru, line 1: I think the path to this file also has to be included in `tests/test_notebooks.py`

tests/e2e/plugin_workflow.ipynb
pmKili: yes, you're right

tests/e2e/plugin_workflow.ipynb
pmKili: done

tests/services/plugins/plugin_folder/sub_folder/helpers.py
pmKili: yes good point

p-desaintchamas (approved):

PR #966, opened 2023-01-31
fix: data should not be imported if the license does not allow it

PierreLeveau:


theodu (changes requested):
it lacks the text case, under the 
`if data_type == TextDataType.LOCAL_FILE:` l82 of `src/kili/services/asset_import/text.py`

hugo-mailfait-kili:
IMO the fix should not be all about the SDK but also the backend, since we want this to be applied for any version of the SDK

Jonas1312 (approved):

PR #967, opened 2023-01-31
ci: run tests against proper server

Jonas1312:


theodu:
I also add @pmKili as a reviewer because he suggested this improvement to me today

.github/scripts/get_test_server.sh
```diff
@@ -0,0 +1,36 @@
...
+    IFS=. read -r major minor patch <<< "$sdk_version"
+    sdk_version_major_minor="$major.$minor"
+
+    prod_version=$(curl --silent --max-time 5 "https://cloud.kili-technology.com/api/label/v2/version" | jq -r .version)
+    echo "prod_version: $prod_version"
+    IFS=. read -r major minor patch <<< "$prod_version"
+    prod_version_major_minor="$major.$minor"
+
+    preprod_version=$(curl --silent --max-time 5 "https://preproduction.cloud.kili-technology.com/api/label/v2/version" | jq -r .version)
+    echo "preprod_version: $preprod_version"
```
theodu, line 20: Does getting a version from this URL guaranty that preproduction is up and working? Just wondering the behavior if there is a deployment issue or another one

.github/scripts/get_test_server.sh
```diff
@@ -0,0 +1,36 @@
...
+        echo "KILI_API_ENDPOINT=https://preproduction.cloud.kili-technology.com/api/label/v2/graphql;"
+    elif [[ $sdk_version_major_minor = $prod_version_major_minor ]]; then
+        echo "TEST_AGAINST=PROD;"
+        echo "KILI_API_ENDPOINT=https://cloud.kili-technology.com/api/label/v2/graphql;"
+    else
+        echo "ERROR: SDK version $sdk_version_major_minor does not match neither preprod version $preprod_version nor prod version $prod_version"
+    fi
+else
+    echo "TEST_AGAINST=STAGING;"
+    echo "KILI_API_ENDPOINT=https://staging.cloud.kili-technology.com/api/label/v2/graphql;"
```
theodu, line 35: just a comment: this is not compatible with our wish to decorrelate the SDK release and the app release. This will need to be updated in the coming quarter

.github/scripts/get_test_server.sh
Jonas1312: Well no not sure :/  But I guess if preprod is up (but not working), many tests later on should crash?

.github/workflows/deploy_doc.yml
```diff
@@ -53,7 +53,6 @@ jobs:
 
       - name: Deploy doc
         run: |
-          pip install -e .
```
theodu, line 56: we had an issue in the past when we didn't prior install the package in the ci.

Jonas1312:
@kili-technology/ml we can even remove the env vars (api key, etc) for the unit tests? They shouldn't be required

.github/workflows/ci.yml
```diff
@@ -50,14 +50,29 @@ jobs:
...
+            echo $test_server_stdout
+            exit 1
+          fi
+
+          test_against=$(echo $test_server_stdout | grep -o "TEST_AGAINST=.*;" | awk -F "=" '{print $2}' | awk -F ";" '{print $1}')
+          echo "TEST_AGAINST=$test_against" >> $GITHUB_ENV
+
+          api_endpoint=$(echo $test_server_stdout | grep -o "KILI_API_ENDPOINT=.*;" | awk -F "=" '{print $2}' | awk -F ";" '{print $1}')
+          echo "KILI_API_ENDPOINT=$api_endpoint" >> $GITHUB_ENV
+
```
theodu, line 67: why do we run the script for unit tests?

.github/workflows/deploy_doc.yml
Jonas1312: it is installed above already

.github/workflows/ci.yml
Jonas1312: yes that's what I just noticed, I don't know why we used to pass the env vars. I'll remove that

.github/scripts/get_test_server.sh
Jonas1312: Yes right. I don't know how to tackle this for now 🤔

pmKili: @theodu why do you think that? Even if we decorrelate the SDK from the app's releases, the SDK will still have to be compatible with all environments. However, if we decorrelate the SDK, we should run E2E tests on LTS environment as well to be sure it remains compatible with old versions from the backend.

pmKili (approved):

Jonas1312:
I updated the code to test on:
- if release branch:
    - preprod if new sdk minor release
    - prod if sdk patch
- else staging

PierreLeveau (approved):

Jonas1312:
@PierreLeveau we should create the API keys before merging. 

I will create the API keys tomorrow and we'll put that in the secrets

Jonas1312:
@PierreLeveau removed test against prod. We can merge once the API keys are put in the github secrets

PR #971, opened 2023-02-01
feat: overload queries for infered return type

theodu:
- feat: overload user-facing queries and make inner queries return a generator
- fix: change internal code for query generator return type
- feat: add test for queries return type
- fix: resolve circular import


theodu:
3 big changes:
- `as_generator` argument is now a keyword argument, providing a value to `as_generator` should explicitly be given by giving the `as_generator` argument. In the function, keyword arguments are after the `*`
- all graphQL queries now return a generator, the transformation into a list must be done after getting the generator result.
- all user queries are overloaded to get the true return type inferred based on the value of `as_generator`

src/kili/queries/asset/media_downloader.py
```diff
@@ -26,7 +28,17 @@ def get_download_assets_function(
...
+    projects = list(
+        ProjectQuery(kili.auth.client)(
+            ProjectWhere(project_id=project_id), fields, QueryOptions(disable_tqdm=True)
+        )
+    )
+    if len(projects) == 0:
+        raise NotFound(
+            f"project ID: {project_id}. Maybe your KILI_API_KEY does not belong to a member of the"
+            " project."
+        )
```
Jonas1312, line 40: aren't we duplicating the code of get_project in src/kili/services/project/__init__.py ?

Jonas1312 (approved):
👍

src/kili/authentication.py
```diff
@@ -112,12 +112,14 @@ def check_expiry_of_key_is_close(self, api_key):
...
-            options=QueryOptions(disable_tqdm=True),
+        api_keys = list(
+            APIKeyQuery(self.client)(
+                fields=["createdAt"],
+                where=APIKeyWhere(api_key=api_key),
+                options=QueryOptions(disable_tqdm=True),
+            )
         )
-        key_creation = datetime.strptime(list(key_object)[0]["createdAt"], "%Y-%m-%dT%H:%M:%S.%fZ")
+        key_creation = datetime.strptime(api_keys[0]["createdAt"], "%Y-%m-%dT%H:%M:%S.%fZ")
```
PierreLeveau, line 122: alternatively, `next(api_keys)` when `api_keys` is a generator also works.

src/kili/queries/asset/media_downloader.py
PierreLeveau: we should not call a service from a core Class (if we do, this should be changed)

PierreLeveau (approved):

src/kili/queries/asset/media_downloader.py
theodu: yes, I duplicated the code but because we had a circular import.
I had it several time for small modifications so I think that It is better to not call the service layer here. Or we can take a ticket to better decorrelate the media_downloader, the services and the queries

src/kili/authentication.py
theodu: comment taken

theodu:
@PierreLeveau I deleted the tests on the kili.get_project because they were failing and because we don't plan to release this feature

PierreLeveau (approved):

PR #976, opened 2023-02-01
ci: update and fix release CIs

Jonas1312:
Changes:
- updated workflow dependencies
- removed entrypoint.sh. Common bash functions are now in utils.sh
- removed release_branch.yml, release_draft_minor.yml, and release_draft_patch.yml.
- instead, use create_draft_release.yml and bump_commit_release_branch,yml:
  - bump_commit_release_branch.yml will have to be triggered manually. It will create a PR with the bump commit. We cannot push directly to the release branch since it is protected.
  - create_draft_release.yml will check out on the release branch, push a tag and create the draft release

TODO:
- test all bash lines in workflows:
  - [x] utils.sh
  - [x] deploy_doc.yml
  - [x] publish.yml
  - [x] bump_commit_release_branch.yml
  - [x] create_draft_release.yml
- test workflows, when possible
  - [ ] publish.yml : cannot test since can't trigger it manually
  - [x] deploy_doc.yml called manually: https://github.com/kili-technology/kili-python-sdk/actions/runs/4084706562
  - [ ] deploy_doc.yml called from publish.yml : cannot test since cannot run publish.yml
  - [ ] bump_commit_release_branch.yml : cannot test since new workflow
  - [ ] create_draft_release.yml : cannot test since new worflow


.github/scripts/utils.sh
```diff
@@ -0,0 +1,31 @@
...
+        src/kili/__init__.py \
+        | grep new_version | sed -r s,"^.*=",,)
+
+    echo "$new_version"
+}
+
+# Takes the version to convert to int
+# 2.129.0 -> 21290
+version_to_int() {
+    echo "$@" | awk -F. '{ printf("%d%03d%d\n", $1,$2,$3); }';
```
PierreLeveau, line 25: `%03d%03d%03d` to be more robust

PierreLeveau (changes requested):

PR #981, opened 2023-02-02
fix: fix bugs in ProjectCopier and MedialDownloader

Jonas1312:
the ProjectQuery was using fields for AssetsQuery

the assets copy in ProjectCopier was not done

e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4075049946

### DIFF
--- a/src/kili/queries/asset/media_downloader.py
+++ b/src/kili/queries/asset/media_downloader.py
@@ -30,7 +30,7 @@ def get_download_assets_function(
         return None
     projects = list(
         ProjectQuery(kili.auth.client)(
-            ProjectWhere(project_id=project_id), fields, QueryOptions(disable_tqdm=True)
+            ProjectWhere(project_id=project_id), ["inputType"], QueryOptions(disable_tqdm=True)
         )
     )
     if len(projects) == 0:
--- a/src/kili/queries/asset/media_downloader.py
+++ b/src/kili/queries/asset/media_downloader.py
@@ -21,13 +21,16 @@ from .exceptions import MissingPropertyError
 
 def get_download_assets_function(
     kili, download_media: bool, fields: List[str], project_id: str, local_media_dir: Optional[str]
-) -> Optional[Callable]:
-    """
-    Return the function to be called after each batch of asset query.
-    It is either None or MediaDownloader's download_assets.
+) -> Tuple[Optional[Callable], List[str]]:
+    """Get the function to be called after each batch of asset query.
+
+    The function is either None or MediaDownloader.download_assets().
+
+    Also returns the fields to be queried, which may be modified
+    if the jsonContent field is necessary.
     """
     if not download_media:
-        return None
+        return None, fields
     projects = list(
         ProjectQuery(kili.auth.client)(
             ProjectWhere(project_id=project_id), ["inputType"], QueryOptions(disable_tqdm=True)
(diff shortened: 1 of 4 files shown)


src/kili/queries/asset/media_downloader.py
```diff
@@ -30,7 +30,7 @@ def get_download_assets_function(
         return None
     projects = list(
         ProjectQuery(kili.auth.client)(
-            ProjectWhere(project_id=project_id), fields, QueryOptions(disable_tqdm=True)
+            ProjectWhere(project_id=project_id), ["inputType"], QueryOptions(disable_tqdm=True)
```
PierreLeveau, line 33: this means that `fields` should be removed in line 44?

src/kili/services/copy_project/__init__.py
```diff
@@ -185,7 +184,12 @@ def download_and_upload_assets(assets):
                 downloaded_assets = self._download_assets(from_project_id, fields, tmp_dir, assets)
                 return self._upload_assets(new_project_id, downloaded_assets)
 
-        return AssetQuery(self.kili.auth.client)(where, fields, options, download_and_upload_assets)
+        asset_gen = AssetQuery(self.kili.auth.client)(
```
PierreLeveau, line 187: not returning AssetQuery does not break anything ?

src/kili/queries/asset/media_downloader.py
Jonas1312: no, this fields is for querying the assets (content, jsoncontent, externalid, etc)

src/kili/services/copy_project/__init__.py
Jonas1312: we don't use the result of _copy_assets():

```python
        if copy_assets:
            logger.info("Copying assets...")
            self._copy_assets(from_project_id, new_project_id)
```

theodu: otherwise you can directly write `asset = list(AssetQuery(self.kili.auth.client)(where, fields, options, download_and_upload_assets))`

theodu (approved):
Thank you, I introduced this bug with the PR on overloading function, sorry!

theodu (changes requested):
just a little comment to have the clean code: instead of iterating on the generator, just call he list function on the generator

Jonas1312:
@theodu yes it works, but it will just waste memory for nothing since we don't use that list later on

theodu (approved):

src/kili/queries/asset/media_downloader.py
```diff
@@ -41,14 +44,17 @@ def get_download_assets_function(
     input_type = projects[0]["inputType"]
     jsoncontent_field_added = False
     if input_type in ("TEXT", "VIDEO") and "jsonContent" not in fields:
-        fields.append("jsonContent")
+        fields = fields + ["jsonContent"]
```
PierreLeveau, line 47: or `fields = [*fields, "jsonContent"]` but this should work.

theodu:
@Jonas1312 indeed, I validate

PierreLeveau (approved):

Jonas1312:
@theodu there are actually no "pythonic" way to consume a generator... https://stackoverflow.com/questions/47456631/simpler-way-to-run-a-generator-function-without-caring-about-items

PR #988, opened 2023-02-06
fix: add delay after mutations

Jonas1312:
I looked at past failed e2e tests and added:

- `append_many_to_dataset`: verify that last batch is uploaded in `BaseBatchImporter.import_batch` by querying assets every second for 5s
- `create_project`: check that project is created every second during 5s
- `send_back_to_queue`, `add_to_review` and `delete_many_from_dataset`: add a callback in `_mutate_from_paginated_call` that checks the last batch

For other asset mutations, `update_properties_in_assets` and `change_asset_external_ids`, I cannot query the assets since the project id is not available in those methods, but required in `AssetWhere`. The simplest thing we can do is add a small delay at the end of `_mutate_from_paginated_call` I think

### DIFF
--- a/src/kili/mutations/asset/__init__.py
+++ b/src/kili/mutations/asset/__init__.py
@@ -5,10 +5,10 @@ Asset mutations
 import warnings
 from typing import Any, Dict, List, Optional, Union
 
-from tenacity import Retrying
+from tenacity import retry
 from tenacity.retry import retry_if_exception_type
 from tenacity.stop import stop_after_delay
-from tenacity.wait import wait_random
+from tenacity.wait import wait_fixed
 from typeguard import typechecked
 
 from kili.authentication import KiliAuth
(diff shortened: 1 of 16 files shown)


src/kili/mutations/asset/__init__.py
```diff
@@ -336,8 +341,30 @@ def delete_many_from_dataset(
         def generate_variables(batch):
             return {"where": {"idIn": batch["asset_ids"]}}
 
+        def verify_last_batch(last_batch: Dict, results: List):
+            """Check that all assets in the last batch have been deleted."""
+            for attempt in Retrying(
```
PierreLeveau, line 346: the decorator syntax is a bit more compact

src/kili/mutations/asset/__init__.py
```diff
@@ -336,8 +341,30 @@ def delete_many_from_dataset(
...
+                stop=stop_after_delay(5),
+                wait=wait_random(min=1, max=2),
```
PierreLeveau, line 348: why do we want a random time?

PierreLeveau (commented):
LGTM but 2 questions

PierreLeveau (approved):

theodu (approved):

PR #990, opened 2023-02-06
feat: add modelName field when querying inference or prediction labels

Jonas1312:
when user queries inference or prediction labels (or both) only, we add the modelName

I'm not really in favor of removing secondstolabel field, since it could break some user code


theodu:
@PierreLeveau didn't we say that putting a .prediction() method was a better idea?

Jonas1312:
@kili-technology/ml is this what we want? It is going to add a lot of LoCs (not to mention the `@overload`)

Jonas1312:
@theodu can you check the pyright error? It's linked to overloading I guess? I'm not sure how to fix that... Thanks

src/kili/queries/label/__init__.py
```diff
@@ -216,6 +216,294 @@ def labels(
...
+        disable_tqdm: bool = False,
+        category_search: Optional[str] = None,
+        *,
+        as_generator: bool = False,
+    ) -> Iterable[Dict]:
+        # pylint: disable=line-too-long
+        """Get prediction labels from a project based on a set of criteria.
+
+        This method is equivalent to the `kili.labels()` method, but it only returns label of type "PREDICTION"
+        made by a model (pre-annotations or real predictions).
```
theodu, line 299: I would remove: `made by a model (pre-annotations or real predictions).`

src/kili/queries/label/__init__.py
```diff
@@ -216,6 +216,294 @@ def labels(
...
+        disable_tqdm: bool = False,
+        category_search: Optional[str] = None,
+        *,
+        as_generator: bool = False,
+    ) -> Iterable[Dict]:
+        # pylint: disable=line-too-long
+        """Get inference labels from a project based on a set of criteria.
+
+        This method is equivalent to the `kili.labels()` method, but it only returns label of type "INFERENCE"
+        made by a model (pre-annotations or real predictions).
```
theodu, line 443: same here

theodu (commented):
one comment in the docstrings

PierreLeveau (approved):

theodu (approved):

PR #997, opened 2023-02-07
fix: e2e test on plugins

pmKili:


gheorghetutunaru (approved):

Jonas1312:
e2e tests still don't pass: https://github.com/kili-technology/kili-python-sdk/actions/runs/4116672545/jobs/7107005673

pmKili:
@Jonas1312 again the problem about credentials in Google. The problem seems to occur randomly. Gregory is off this week so we cannot do many things

Jonas1312 (approved):

PR #998, opened 2023-02-08
feat: add `wait_until_availability` argument for `append_many_to_dataset`

Jonas1312:
e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4133821802

still one plugin notebook failing

### DIFF
--- a/src/kili/helpers.py
+++ b/src/kili/helpers.py
@@ -14,6 +14,7 @@ from typing import Callable, Dict, List, Optional, Type, TypeVar, Union
 
 import pyparsing as pp
 import requests
+import tenacity
 from typing_extensions import get_args, get_origin
 
 from kili.constants import mime_extensions_for_IV2
@@ -346,3 +347,31 @@ def disable_tqdm_if_as_generator(as_generator: bool, disable_tqdm: bool):
             " generator return type"
         )
     return disable_tqdm
+
+
+class RetryLongWaitWarner:  # pylint: disable=too-few-public-methods
+    """Class that warns when retry takes too long."""
+
+    def __init__(
+        self, method_name: str, warn_after: float = 10, warn_message: Optional[str] = None
+    ) -> None:
+        """Class that warns when retry takes too long.
+
+        Args:
+            method_name: method name to be used in the warning message.
+            warn_after: time in seconds after which the warning is raised.
+            warn_message: custom warning message. If not provided, a default message is used.
+        """
+        self.method_name = method_name
+        self.warn_after = warn_after
+        self.warn_message: str = warn_message or (
+            f"Warning: '{self.method_name}' is taking a long time to complete. This may be due"
+            " to a large number of mutations to be processed by the server."
+        )
+
+        self.warned = False
+
+    def __call__(self, retry_state: tenacity.RetryCallState):
+        if not self.warned and float(retry_state.outcome_timestamp or 0) > self.warn_after:
+            warnings.warn(self.warn_message, category=RuntimeWarning)
+            self.warned = True
--- a/src/kili/helpers.py
+++ b/src/kili/helpers.py
@@ -353,25 +353,25 @@ class RetryLongWaitWarner:  # pylint: disable=too-few-public-methods
     """Class that warns when retry takes too long."""
 
     def __init__(
-        self, method_name: str, warn_after: float = 10, warn_message: Optional[str] = None
+        self,
+        warn_message: str,
+        logger_func: Callable,
+        warn_after: float = 10,
     ) -> None:
         """Class that warns when retry takes too long.
 
         Args:
-            method_name: method name to be used in the warning message.
-            warn_after: time in seconds after which the warning is raised.
             warn_message: custom warning message. If not provided, a default message is used.
+            logger_func: function to log the message (print, warning, logger.warning, etc.)
+            warn_after: time in seconds after which the warning is raised.
         """
-        self.method_name = method_name
+        self.warn_message = warn_message
+        self.logger_func = logger_func
         self.warn_after = warn_after
-        self.warn_message: str = warn_message or (
-            f"Warning: '{self.method_name}' is taking a long time to complete. This may be due"
-            " to a large number of mutations to be processed by the server."
-        )
 
         self.warned = False
 
     def __call__(self, retry_state: tenacity.RetryCallState):
         if not self.warned and float(retry_state.outcome_timestamp or 0) > self.warn_after:
-            warnings.warn(self.warn_message, category=RuntimeWarning)
+            self.logger_func(self.warn_message)
             self.warned = True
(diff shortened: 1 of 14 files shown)


src/kili/helpers.py
```diff
@@ -346,3 +347,31 @@ def disable_tqdm_if_as_generator(as_generator: bool, disable_tqdm: bool):
...
+        Args:
+            method_name: method name to be used in the warning message.
+            warn_after: time in seconds after which the warning is raised.
+            warn_message: custom warning message. If not provided, a default message is used.
+        """
+        self.method_name = method_name
+        self.warn_after = warn_after
+        self.warn_message: str = warn_message or (
+            f"Warning: '{self.method_name}' is taking a long time to complete. This may be due"
+            " to a large number of mutations to be processed by the server."
```
theodu, line 369: For the case of the asynchronous processes (like frame video import of deleted assets), we should instead find another message.
Otherwise, it looks like a problem of lack of resources whereas it is intended and is a feature to be asynchronous.

theodu: If mutations can take some times, it could be be 2 differents reasons:
- intended asynchronous process
- db being overwhelmed (the topic of your ticket)

For me it should be handled differently (1st case is better logging for the user, 2nd case is your ticket to fix the issue). Here I have the impression that we code the same solution for the 2 differents issues

src/kili/mutations/asset/__init__.py
```diff
@@ -93,6 +94,7 @@ def append_many_to_dataset(
                 - For VIDEO projects (and not VIDEO_LEGACY), you can specify a value with key 'processingParameters' to specify the sampling rate (default: 30).
                     Example for one asset: `json_metadata_array = [{'processingParameters': {'framesPlayedPerSecond': 10}}]`.
             disable_tqdm: If `True`, the progress bar will be disabled
+            blocking: If `True`, the function will wait until the assets are imported.
```
theodu, line 97: For me, we could find another name more explicit like `wait_until_sucess`? And add a bit more info on the arg description
Here I doubt that someone without the context of all our discussion would understand when to use this when having a look at the doc

src/kili/mutations/asset/__init__.py
```diff
@@ -58,6 +58,7 @@ def append_many_to_dataset(
         json_content_array: Optional[List[List[Union[dict, str]]]] = None,
         json_metadata_array: Optional[List[dict]] = None,
         disable_tqdm: bool = False,
+        blocking: bool = True,
```
theodu, line 61: shouldn't we force it to be a keyword argument?

src/kili/helpers.py
Jonas1312: I agree for the two reasons why an import might take some time.

But I think it should be handled the same way (with the same code, but different warnings or logging). The user is not supposed to know that some assets are imported asynchronously or synchronously in the kili service import_assets.

From a user point of view:
- if I use kili.append_many_to_dataset(wait_until_sucess=False), I deal with the consequences, i.e. I shouldn't query just after it
- if I use kili.append_many_to_dataset(wait_until_sucess=True), I am 100% sure that my import is finished and visible in kili.assets()

However, we can indeed change the warning message:
- for sync assets: we **logger.warn** it's taking some time because of big asset
- for async assets: we **logger.info** it's taking some time because the backend needs to process the file and it's totally normal

@kili-technology/ml let me know if it's good for you

theodu: I agree with you that the wait_until_success should handle the two cases and that it should just be different logging messages.

src/kili/mutations/asset/__init__.py
```diff
@@ -322,13 +330,17 @@ def delete_many_from_dataset(
...
+        wait_until_sucess: Optional[bool] = True,
     ) -> Asset:
         """Delete assets from a project.
 
         Args:
             asset_ids: The list of asset internal IDs to delete.
             external_ids: The list of asset external IDs to delete.
             project_id: The project ID. Only required if `external_ids` argument is provided.
+            wait_until_sucess: If True, the method will wait until the assets are fully deleted (synchronous mode).
+                If False, the method will return faster but the assets may not be fully deleted (asynchronous mode).
```
theodu, line 344: I did the test on the SDK and it seems that even for asynchronous processes in the app, the SDK still wait until the end of the process to give the hand back

theodu (changes requested):
changes seen with you in a call.
Make a distinction for resolvers do a process in the backend from the one that executes all the process in a synchronous time (modulo the replica lag)

Jonas1312:
e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4136415367

PierreLeveau (approved):

Jonas1312:
e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4141544070

src/kili/mutations/asset/__init__.py
```diff
@@ -342,9 +348,9 @@ def generate_variables(batch):
             return {"where": {"idIn": batch["asset_ids"]}}
 
         @retry(
-            stop=stop_after_delay(5),
             wait=wait_fixed(1),
```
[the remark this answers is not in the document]
theodu: I think that we should check with @utix (aurélien) and @HNicolas to validate such a decision.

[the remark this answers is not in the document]
Jonas1312: I agree with you indeed. We can use an exponential wait, we start at 1 sec and double the delay at every retry with a max wait of 16s. Because I don't think we can do better SDK side for now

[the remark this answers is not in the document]
Jonas1312: Also, can't we check the last asset of the last batch only? Instead of checking the whole batch? I'm not sure if it's processed in parallel in the backend or not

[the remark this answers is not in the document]
theodu: Ah yeah, good point, I don't know how it is handled exactly, we could have a look at the label-import service or the appendManyFramesToDatasetAsynchronously.
But also, it is maybe possible in some cases to have asynchronous batches between synchronous ones. If we test the last batch, we wouldn't catch the asynchronous behavior of the in-between batches. Might be worth to check if all asynchronous batches are done last in the SDK ?

PierreLeveau (approved):

Jonas1312:
e2e ongoing: https://github.com/kili-technology/kili-python-sdk/actions/runs/4161283118

src/kili/mutations/asset/__init__.py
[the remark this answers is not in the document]
Jonas1312: I added indefinite retries with exponential delay (min=1, max=8) and query only last element of last batch of the mutation

Jonas1312:
e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4161535151

theodu (approved):

Jonas1312:
e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4161927566

notebook plugins error, it's "normal"

I merge

PR #1000, opened 2023-02-10
refactor: change GraphQL client for gql

Jonas1312:


Jonas1312:
we can start to review this, I will add the subscriptionGqlClient later since it's  a bit complicated

e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4164362642

Jonas1312:
e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4164862271

PierreLeveau (changes requested):

src/kili/exceptions.py
```diff
@@ -1,6 +1,12 @@
 """Exceptions of the package."""
 from typing import Optional
 
+import graphql
+from gql.transport import exceptions
+
+TransportQueryError = exceptions.TransportQueryError
```
PierreLeveau, line 7: to document more accurately I would rather write that:
```python
class TransportQueryError(exceptions.TransportQueryError):
     """Raised when ... """
     pass

class GraphQLError(graphql.GraphQLError):
     """"Raised when ... """
    pass
```

Jonas1312: I can't do that because the exceptions will be raised by graphql and gql libraries.

If I test for `pytest.raises(KiliSdkGraphQLError)`, the test will fail because `isinstance(graphql.GraphQLError(), KiliSdkGraphQLError)` is false

src/kili/graphql/operations/api_key/queries.py
```diff
@@ -38,9 +36,6 @@ def graphql_where_builder(self):
 class APIKeyQuery(GraphQLQuery):
     """ProjectUser query."""
 
-    FORMAT_TYPE = ApiKey
```
theodu, line 41: you removed both of them

src/kili/utils/pagination.py
```diff
@@ -186,10 +186,13 @@ def generate_variables(batched_properties):
     batch = None
     for batch_number, batch in enumerate(batch_object_builder(properties_to_batch, batch_size)):
         payload = generate_variables(batch)
-        result = api_throttle(self.auth.client.execute)(request, payload)
+        try:
+            result = api_throttle(self.auth.client.execute)(request, payload)
+        except (TransportQueryError, GraphQLError) as err:
+            raise ValueError(
```
theodu, line 192: Why don't we rethrow a TransportQueryError here?
it is one that we send when we don't execute mutations by batch isn't it?

tests/e2e/test_graphql_client.py
```diff
@@ -0,0 +1,52 @@
...
+
+    kili.auth.client._gql_client.schema = None
+    kili.auth.client._gql_client.fetch_schema_from_transport = False
+    mocked_validate = mock.MagicMock()
+    kili.auth.client._gql_client.validate = mocked_validate
+
+    with pytest.raises(TransportQueryError, match="Cannot query field"):
+        kili.auth.client.execute(query)
+
+    mocked_validate.assert_not_called()
```
theodu, line 52: What is the difference between the 2 cases ?

src/kili/graphql/__init__.py
```diff
@@ -149,43 +145,36 @@ def execute_query_from_paginated_call(
...
-            except KeyError as exception:
-                raise NonExistingFieldError(
-                    f"Cannot query field {field} on object {typed_dict_class.__name__}. Admissible"
-                    " fields are: \n- "
-                    + "\n- ".join(type_of_fields.keys())
-                ) from exception
-            if isinstance(field, str):
-                fragment += f" {field}"
-            else:
-                raise TypeError("Please provide the fields to query as strings")
```
theodu, line 190: So all this feature of rejecting wrong fields is done by gql now? What message do I get if I ask for asset.do_not_exist for example ?

theodu (commented):
.

tests/e2e/test_graphql_client.py
Jonas1312: one is local query validation, one is remote query validation by the backend

src/kili/utils/pagination.py
Jonas1312: self.auth.client.execute could fail for GraphQLError if the graphql query is invalid according to the schema

or can fail for TransportQueryError if the query is refused by the backend

Jonas1312: I raise a value error but the `from err` will also add the original error (TransportQueryError or GraphQLError)

src/kili/graphql/__init__.py
Jonas1312: yes, the query will be refused if the field is invalid

```text
graphql.error.graphql_error.GraphQLError: Cannot query field 'sdfsdfsdfsdfsdfsdfs' on type 'Asset'.

GraphQL request:4:21
3 |             data: assets(where: $where, skip: $skip, first: $first) {
4 |                  id sdfsdfsdfsdfsdfsdfs
  |                     ^
5 |             }
```

tests/e2e/test_graphql_client.py
theodu: I don't get the difference between the two. When do we do a local validation or backend validation?
Can we choose when to validate locally or in the backend? And are there reasons to rather do it locally or in the backend?

Jonas1312: local validation checks for correct query format

backend validation checks for query format and other rights (cannot query assets if no project id, etc.)

Jonas1312: we can choose not to do it locally yes

we can do it locally since it doesn't cost anything more to do it locally

can be used for testing purposes

theodu: Thanks!

theodu (approved):

Jonas1312:
e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4182144096

tests/e2e/test_graphql_client.py
PierreLeveau: + the response time should be faster if it is first done locally

Jonas1312:
e2e ongoing: https://github.com/kili-technology/kili-python-sdk/actions/runs/4193060622

PR #1029, opened 2023-02-20
feat: add `kili.add_data_connection`

Jonas1312:


docs/sdk/cloud_storage.md
```diff
@@ -3,3 +3,7 @@
 ## Queries
 
 ::: kili.queries.data_integration.__init__.QueriesDataIntegration
+
+## Mutations
+
+:::kili.mutations.data_connection.__init__.MutationsDataConnection
```
theodu, line 9: I think that we could also add a little paragraph to explain what is the difference between what we call a data connection and a data integration.
And to to explain that the integration can only be done in the front currently and link to https://docs.kili-technology.com/docs/adding-assets-to-project#creating-a-remote-storage-integration
also @PierreLeveau in the readMe doc, it is called "Remote Storage"

src/kili/mutations/data_connection/queries.py
```diff
@@ -0,0 +1,11 @@
+"""
+Queries of data connection mutations
+"""
+
+GQL_ADD_PROJECT_DATA_CONNECTION = """
+mutation ($data: DataConnectionInput!) {
```
theodu, line 6:
```suggestion
mutation addDataConnection($data: DataConnectionInput!) {
```

So in apollo studio, we can see the name of the call `addDataConnection` instead of seeing the stringified query. Much easier to read

theodu (changes requested):

docs/sdk/cloud_storage.md
Jonas1312: Yes I'm going to add some more info in the doc

Jonas1312: @kili-technology/ml we rename all "cloud storage" to "remote storage"?

docs/sdk/cloud_storage.md
```diff
@@ -1,5 +1,18 @@
 # Cloud storage module
 
+## Data Integration versus Data Connection
+
+A data integration is a connection between a Kili organization and a remote/cloud storage (AWS, GCP or Azure).
+Once a data integration is created, it can be used in any project of the organization.
+More information about how to create a data integration can be found [here](https://docs.kili-technology.com/docs/adding-assets-to-project#creating-a-remote-storage-integration).
```
theodu, line 7:
```suggestion
Adding a data integration from the API is currently not supported. More information about how to create a data integration can be found [here](https://docs.kili-technology.com/docs/adding-assets-to-project#creating-a-remote-storage-integration).
```

theodu (changes requested):

theodu (approved):

docs/sdk/cloud_storage.md
PierreLeveau: not a big deal, we will rename it when the renaming will be effective.

PierreLeveau (approved):

PR #1033, opened 2023-02-20
feat: cache the graphql schema in graphql client

Jonas1312:


theodu (approved):

src/kili/graphql/graphql_client.py
```diff
@@ -60,7 +63,59 @@ def __init__(
...
+        with graphql_schema_path.open("w", encoding="utf-8") as file:
+            file.write(schema_str)
+
+    def _get_graphql_schema_path(self) -> Path:
+        """
+        Get the path of the GraphQL schema
+        """
+        endpoint_netloc = urlparse(self.endpoint).netloc
+        version = self.get_kili_app_version()
+        timestamp = datetime.now().strftime(r"%Y%m%d")  # cache for one day at most
```
PierreLeveau, line 104: it won't be a 1 day cache if it is generated at 23:59:59. It would be better to index the cache by the timestamp in seconds, and check if the cache is valid depending on this timestamp being more than 1 day-old.

src/kili/graphql/graphql_client.py
```diff
@@ -60,7 +63,59 @@ def __init__(
...
+
+        filename = f"{endpoint_netloc}_{version}_{timestamp}.graphql"
+        dir_ = Path.home() / ".cache" / "kili" / "graphql"
+
+        return dir_ / filename
+
+    def get_kili_app_version(self) -> str:
```
PierreLeveau, line 111: if the cache is indexed by the version, no need to check the timestamp after all ?

PierreLeveau (changes requested):

Jonas1312:
e2e ongoing: https://github.com/kili-technology/kili-python-sdk/actions/runs/4232222572/jobs/7351689099

src/kili/graphql/graphql_client.py
Jonas1312: yes I only keep the endpoint name and app version. The schema shouldn't change anyway for the same version

PierreLeveau (approved):

PR #1036, opened 2023-02-22
fix: fix coco bbox format

theodu:


theodu:
COCO format: https://cocodataset.org/#format-data

> 1. Object Detection
Each object instance annotation contains a series of fields, including the category id and segmentation mask of the object. The segmentation format depends on whether the instance represents a single object (iscrowd=0 in which case polygons are used) or a collection of objects (iscrowd=1 in which case RLE is used). Note that a single object (iscrowd=0) may require multiple polygons, for example if occluded. Crowd annotations (iscrowd=1) are used to label large groups of objects (e.g. a crowd of people). In addition, an enclosing bounding box is provided for each object (box coordinates are measured from the top left image corner and are 0-indexed). Finally, the categories field of the annotation structure stores the mapping of category id to category and supercategory names. See also the [detection](https://cocodataset.org/#detection-2020) task.

Jonas1312 (approved):

src/kili/services/export/format/coco/__init__.py
```diff
@@ -313,7 +315,7 @@ def _get_coco_image_annotations(
                 id=annotation_j,
                 image_id=asset["id"],
                 category_id=cat_kili_id_to_coco_id[categories[0]["name"]],
-                bbox=[int(np.min(p_x)), int(np.min(p_y)), int(np.max(p_x)), int(np.max(p_y))],
+                bbox=[int(x_min), int(y_min), int(_w_), int(_h_)],
```
PierreLeveau, line 318: to make sure this does not happen again, can we add a unit test? if can be worth it to extract a method `_get_coco_bbox_from_kili_bpoly`

PierreLeveau (changes requested):

PierreLeveau (approved):

PR #1037, opened 2023-02-22
docs: ✏️ Detail the fields available for label in plugin handler

gheorghetutunaru:


src/kili/services/plugins/model.py
```diff
@@ -46,9 +46,12 @@ def on_submit(
         Handler for the submit action, triggered when a default label is submitted into Kili.
 
         Args:
-            label: label submitted to Kili
-            asset_id: id of the asset on which the label was submitted
-
+            label: Label submitted to Kili: a dictionary containing the following fields:
```
theodu, line 49: don't we want to directly link to the graphQL reference page ? https://docs.kili-technology.com/reference/graphql-api#label
Its is the one that we link in the SDK doc to show all fields that the client can query

gheorghetutunaru: The issue is that we don't have all the fields from the Graphql reference, that's why it was confusing to know what is present and what not

theodu: oh okay!

theodu (approved):

florianlega (approved):

PR #1040, opened 2023-02-27
refactor: revert old GraphQLError exception

Jonas1312:
e2e pass: https://github.com/kili-technology/kili-python-sdk/actions/runs/4280730417

### DIFF
--- a/src/kili/exceptions.py
+++ b/src/kili/exceptions.py
@@ -1,11 +1,9 @@
 """Exceptions of the package."""
 from typing import Optional
 
-import graphql
-from gql.transport import exceptions
 
-TransportQueryError = exceptions.TransportQueryError
-GraphQLError = graphql.GraphQLError
+class GraphQLError(Exception):
+    """Raised when the GraphQL call returns an error"""
 
 
 class NotFound(Exception):
--- a/src/kili/exceptions.py
+++ b/src/kili/exceptions.py
@@ -1,9 +1,24 @@
 """Exceptions of the package."""
-from typing import Optional
+from typing import Dict, List, Optional
 
 
 class GraphQLError(Exception):
     """Raised when the GraphQL call returns an error"""
+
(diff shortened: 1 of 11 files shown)


src/kili/exceptions.py
```diff
@@ -1,11 +1,9 @@
 """Exceptions of the package."""
 from typing import Optional
 
-import graphql
-from gql.transport import exceptions
 
-TransportQueryError = exceptions.TransportQueryError
-GraphQLError = graphql.GraphQLError
+class GraphQLError(Exception):
+    """Raised when the GraphQL call returns an error"""
```
theodu, line 7: our class was :
```
class GraphQLError(Exception):
    """
    Used when the GraphQL call returns an error
    """

    def __init__(self, error, batch_number=None):
        if isinstance(error, list):
            error = error[0]
        if isinstance(error, dict) and "message" in error:
            error_msg = error["message"]
        else:
            error_msg = str(error)

        if batch_number is None:
            super().__init__(f'error: "{error_msg}"')
        else:
            super().__init__(f'error at index {100*batch_number}: {error_msg}"')
```

Is it on purpose to drop the __init__ code ?

Jonas1312: the content of the graphqlerror will be printed in the graphql.GraphQLError or the gql.transports.exceptions.Transportqueryerror since we `raise GraphQLError() from err`

theodu: is the message similar? in case clients were doing test with the graphQL error message

Jonas1312: no it's quite different, I'll revert to the previous init code. We'll have the error twice in the traceback but that's not an issue

Jonas1312:
e2e ongoing: https://github.com/kili-technology/kili-python-sdk/actions/runs/4281205075

PierreLeveau (approved):

theodu (approved):

PR #1042, opened 2023-02-27
feat: add more arguments to issues queries

theodu:


Jonas1312 (commented):
we need to update the @overload too I guess

PierreLeveau (approved):

PR #1047, opened 2023-02-28
refactor: rename cloud storage methods and update doc

Jonas1312:
I had to rework test_pagination.py because pyright CI was not passing because of row_generator_from_paginated_calls method

PR #1049, opened 2023-02-28
chore: merge master in release/2.129.0

theodu:


Jonas1312:
we need to merge this one to fix the pyright errors: https://github.com/kili-technology/kili-python-sdk/pull/1047

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.129.3
```
Jonas1312, line 2: we will have to double bump before next release

src/kili/__init__.py
```diff
@@ -5,4 +5,4 @@
 Kili Python SDK
 """
 
-__version__ = "2.129.3"
```
Jonas1312, line 8: same here

Jonas1312 (approved):
just need to bump

setup.cfg
theodu: I just added the commit of bump to restore 2.129.3 on the branch

PR #1072, opened 2023-03-06
chore: remove version mismatch check

Jonas1312:


Jonas1312:
> Maybe we should then get rid of the API Key query? It was the only reason why we kept it in the SDK after removing the mutations I think

what do you mean?  We still want to check that the api key is valid when we init the Kiliauth no?

theodu:
> > Maybe we should then get rid of the API Key query? It was the only reason why we kept it in the SDK after removing the mutations I think
> 
> what do you mean? We still want to check that the api key is valid when we init the Kiliauth no?

ah yeah indeed we do it in the SDK at the client init, I forgot this. I validate!

theodu (approved):

PierreLeveau (approved):

PR #1079, opened 2023-03-08
feat: warn and return if argument of type list is empty

Jonas1312:
add a helperthat raises warning and return bool if the provided input list is empty 

Some methods already check for the len(input_array), so I didn't modify them

e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4382202318

### DIFF
--- a/src/kili/helpers.py
+++ b/src/kili/helpers.py
@@ -400,7 +400,7 @@ def skip_if_empty_arguments(
 
     Args:
         all_non_empty: sequence of argument names that must be all non-empty.
-        any_non_empty: sequence of argument names that must be at least one non-empty.
+        any_non_empty: sequence of argument names that must have at least one non-empty value.
     """
 
     def is_empty_sequence(arg_value) -> bool:
@@ -408,7 +408,7 @@ def skip_if_empty_arguments(
 
     def get_arg_value(arg_name: str, args: Tuple[Any, ...], kwargs: Dict[str, Any], func: Callable):
         """
-        Get the value of an argument `arg_name` from args and kwargs.
+        Get the value of an argument `arg_name` from the args and kwargs of the decorated function.
         """
         if arg_name in kwargs:
             return kwargs[arg_name]
@@ -419,7 +419,7 @@ def skip_if_empty_arguments(
         except IndexError:
             return None
 
-    def get_args_dict(
+    def get_args_as_dict(
         arg_names: Sequence[str], args: Tuple[Any, ...], kwargs: Dict[str, Any], func: Callable
     ) -> Dict[str, Any]:
         args_dict = {}
@@ -436,7 +436,7 @@ def skip_if_empty_arguments(
 
             # we want all arguments to be non-empty
             if all_non_empty:
-                args_dict = get_args_dict(all_non_empty, args, kwargs, func)
+                args_dict = get_args_as_dict(all_non_empty, args, kwargs, func)
 
                 empty_args = [
                     arg_name
@@ -456,7 +456,7 @@ def skip_if_empty_arguments(
 
             # we need at least one argument to be non-empty
             if any_non_empty:
-                args_dict = get_args_dict(any_non_empty, args, kwargs, func)
+                args_dict = get_args_as_dict(any_non_empty, args, kwargs, func)
 
                 if args_dict and all(
                     is_empty_sequence(arg_value) for arg_value in args_dict.values()
--- a/src/kili/helpers.py
+++ b/src/kili/helpers.py
@@ -10,18 +10,7 @@ import os
 import re
 import warnings
 from json import dumps, loads
-from typing import (
-    Any,
-    Callable,
-    Dict,
-    List,
-    Optional,
-    Sequence,
-    Tuple,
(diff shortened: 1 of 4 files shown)


src/kili/mutations/asset/__init__.py
```diff
@@ -124,6 +126,7 @@ def append_many_to_dataset(
 
         if content_array is None and json_content_array is None:
```
PierreLeveau, line 127: This condition can't happen since there is the decorator? there can be similar conditions in the code.

src/kili/helpers.py
```diff
@@ -374,3 +385,96 @@ def __call__(self, retry_state: tenacity.RetryCallState):
...
+                empty_args = [
+                    arg_name
+                    for arg_name, arg_value in args_dict.items()
+                    if is_empty_sequence(arg_value)
+                ]
+
+                if empty_args:
+                    warnings.warn(
+                        (
+                            f"Skipping '{func.__name__}' because the following arguments are"
```
PierreLeveau, line 450: `"Method '{func.__name__}' did nothing because ...`

src/kili/helpers.py
```diff
@@ -374,3 +385,96 @@ def __call__(self, retry_state: tenacity.RetryCallState):
...
+            # we need at least one argument to be non-empty
+            if any_non_empty:
+                args_dict = get_args_as_dict(any_non_empty, args, kwargs, func)
+
+                if args_dict and all(
+                    is_empty_sequence(arg_value) for arg_value in args_dict.values()
+                ):
+                    warnings.warn(
+                        (
+                            f"Skipping '{func.__name__}' because the following arguments are"
```
PierreLeveau, line 466: same as above

src/kili/helpers.py
```diff
@@ -374,3 +385,96 @@ def __call__(self, retry_state: tenacity.RetryCallState):
...
+                            f" empty: {', '.join(empty_args)}."
+                        ),
+                        stacklevel=4,
+                    )
+                    has_warned = True
+
+            # we need at least one argument to be non-empty
+            if any_non_empty:
```
PierreLeveau, line 458: this should run before the all_non_empty test, which should be skipped if any_non_empty has warned.

src/kili/helpers.py
```diff
@@ -374,3 +385,96 @@ def __call__(self, retry_state: tenacity.RetryCallState):
...
+        args_dict = {}
+        for arg_name in arg_names:
+            args_value = get_arg_value(arg_name, args, kwargs, func)
+            if args_value is not None:
+                args_dict[arg_name] = args_value
+        return args_dict
+
+    def decorator(func: Callable[..., D]) -> Callable[..., Optional[D]]:
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs) -> Optional[D]:
```
PierreLeveau, line 434: alternatively this could return an empty object of type D, to not change the return type. But I guess that's more complicated to implement.

PierreLeveau (changes requested):

src/kili/mutations/asset/__init__.py
Jonas1312: the decorator only skips when content_array  and json_content_array are empty lists

src/kili/helpers.py
Jonas1312: why should it be run before?

src/kili/helpers.py
Jonas1312: If we have a method that returns a {"id": "123456"} for example, you would return an empty dict?  I think returning None is fine

theodu (commented):
My global feeling is that it is a huge amount of complicated code for a simple problem.
I guess you have done it that way so it factories the code and it avoids forgetting it but we can still forget to add the decorator.
Shouldn't we only add an early return with a simple condition in each function? It would make less code and would be easier to maintain.

src/kili/helpers.py
PierreLeveau: good point

src/kili/mutations/asset/__init__.py
```diff
@@ -112,6 +113,13 @@ def append_many_to_dataset(
             - For more detailed examples on how to import text assets,
                 see [the recipe](https://github.com/kili-technology/kili-python-sdk/blob/master/recipes/import_text_assets.ipynb).
         """
+        if check_warn_empty_list(
+            "append_many_to_dataset", "content_array", content_array
+        ) or check_warn_empty_list(
+            "append_many_to_dataset", "json_content_array", json_content_array
+        ):
+            return None
+
```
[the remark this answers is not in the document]
Jonas1312: No in python only the first check of the logical OR will be executed

PierreLeveau (approved):

Jonas1312:
ret = kili.add_to_review([], [])

will warn like this:

```text
/Users/jonasm/kili-python-sdk/sandbox.py:25: UserWarning: Method 'add_to_review' did nothing because the following argument is empty: asset_ids.
  ret = kili.add_to_review([], [])
```

theodu (approved):

PR #1108, opened 2023-03-17
fix: merged COCO export working

PierreLeveau:


src/kili/services/export/format/coco/__init__.py
```diff
@@ -96,19 +96,30 @@ def _save_assets_export(
...
+        else:  # merged
+            _convert_kili_semantic_to_coco(
+                jobs=self.project_json_interface["jobs"],
+                assets=assets,
+                output_dir=Path(output_directory) / self.project_id,
+                title=self.project_title,
+                project_input_type=self.project_input_type,
+                annotation_modifier=annotation_modifier,
+                merged=True,
+            )
```
Jonas1312, line 122: for the merged call to _convert_kili_semantic_to_coco, we should still check for compatible jobs I think

src/kili/services/export/format/coco/__init__.py
```diff
@@ -157,113 +168,226 @@ def _convert_kili_semantic_to_coco(
...
-                        cat_kili_id_to_coco_id,
+                        cat_kili_id_to_coco_id[job_name],
                         annotation_offset,
                         coco_image,
                         annotation_modifier,
                     )
                 coco_annotations.extend(coco_img_annotations)
-
+            else:
+                raise NotImplementedError
```
Jonas1312, line 390: why don't we cover the case "is_single_job is False"?

src/kili/services/export/format/coco/__init__.py
```diff
@@ -157,113 +168,226 @@ def _convert_kili_semantic_to_coco(
...
+        label_filenames = []
+        for job_name, job in jobs.items():
+            labels_json["images"], labels_json["annotations"] = _get_coco_images_and_annotations(
+                {job_name: job},
+                assets,
+                cat_kili_id_to_coco_id,
+                project_input_type,
+                annotation_modifier,
+                is_single_job=True,
+            )
```
Jonas1312, line 200: I'm not sure if it's necessary to add this argument "is_single_job". We can just do `is_single_job = len(jobs) == 1` in the methods called later on maybe

Jonas1312 (changes requested):

src/kili/services/export/format/coco/__init__.py
PierreLeveau: the "merged" syntax outputs categories name  "job/category_name", while the "split" syntax outputs "category_name", so I think we should still tell the function which layout it is.

src/kili/services/export/format/coco/__init__.py
PierreLeveau: I'l add that in this PR then

Jonas1312 (approved):

PR #1111, opened 2023-03-20
chore: increase gql logger level

Jonas1312:
gql graphql client logs all queries sent which is too verbose by default

```
INFO:gql.transport.requests:>>> {"query": "query projects($where: ProjectWhere!, $first: PageSize!, $skip: Int!) {\n  data: projects(where: $where, first: $first, skip: $skip) {\n    inputType\n  }\n}", "variables": {"where": {"id": "123456", "searchQuery": null, "shouldRelaunchKpiComputation": null, "updatedAtGte": null, "updatedAtLte": null}, "skip": 0, "first": 1}}
INFO:gql.transport.requests:<<< {"data":{"data":[{"inputType":"IMAGE"}]}}
```

### DIFF
--- a/src/kili/graphql/graphql_client.py
+++ b/src/kili/graphql/graphql_client.py
@@ -3,6 +3,7 @@ GraphQL Client
 """
 
 import json
+import logging
 import random
 import string
 import threading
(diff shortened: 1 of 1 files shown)


Jonas1312:
Hugo confirms it reduced the log's verbosity

PR #1123, opened 2023-03-23
fix: query from external ids improved

PierreLeveau:
Getting all the asset ids and external ids and putting them in a dict should be efficient enough.

### DIFF
--- a/src/kili/services/helpers.py
+++ b/src/kili/services/helpers.py
@@ -6,7 +6,6 @@ from pathlib import Path
 from typing import Any, Dict, Generator, Iterable, List, Optional, TypeVar
 
 from kili.authentication import KiliAuth
-from kili.constants import QUERY_BATCH_SIZE
 from kili.exceptions import NotFound
 from kili.graphql import QueryOptions
 from kili.graphql.operations.asset.queries import AssetQuery, AssetWhere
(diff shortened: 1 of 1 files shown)


src/kili/services/helpers.py
```diff
@@ -91,6 +86,18 @@ def infer_ids_from_external_ids(
...
 
 
+def _build_id_map(auth, asset_external_ids, project_id):
+    assets = AssetQuery(auth.client)(
+        AssetWhere(project_id),
+        ["id", "externalId"],
+        QueryOptions(disable_tqdm=True),
+    )
+    id_map: Dict[str, str] = {}
+    for asset in (asset for asset in assets if asset["externalId"] in set(asset_external_ids)):
```
Jonas1312, line 96: you should put the `set(asset_external_ids)` outside of the generator, because a new set will be computed for each asset in assets

Jonas1312 (approved):
we should just put a comment saying we don't restrict the assetwhere

lgtm otherwise

src/kili/services/helpers.py
PierreLeveau: no it should not be recomputed, see:
```
In [1]: a = 5

In [2]: for b in range(a):
   ...:     print(a)
   ...:     a+=1
   ...:
5
6
7
8
9

```

Jonas1312: I'm not 100% sure honestly.

This is not the same "in" in the code and in your example.

The "in" in for `for b in range(a)` is an "iterator" in.

The "in" in ` ... in set(...)` is a membership test. Everytime it's executed, the set() will be evaluated again:

```python
class MyClass:
    def __init__(self):
        self.a = [1, 2, 3]

    def __call__(self):
        print("class called")
        return self.a

a = MyClass()
for nb in (nb for nb in [1, 2, 3] if nb in set(a())):
    print("for loop: ", nb)
```

prints:

```python
class called
for loop: 1
class called
for loop: 2
class called
for loop: 3
```

PierreLeveau: oh you're right, my mistake

src/kili/services/helpers.py
```diff
@@ -91,6 +89,35 @@ def infer_ids_from_external_ids(
...
+                )
+            )
+        assets = chain(*assets_generators)
+    else:
+        # query all assets
+        assets = AssetQuery(auth.client)(
+            AssetWhere(project_id),
+            ["id", "externalId"],
+            QueryOptions(disable_tqdm=True),
+        )
```
Jonas1312, line 113: When `len(asset_external_ids) <= 1000`, we don't want to pass the `external_id_contains=asset_external_ids` in the AssetWhere? It's smaller than 1000 so the backend should not crash?

I wonder if we should just simply batch everything, regardless of the len(asset_external_ids)

Jonas1312 (approved):

src/kili/services/helpers.py
PierreLeveau: you're right, we can even batch with 1000 size

PierreLeveau: changed

PR #1129, opened 2023-03-29
ci: fix bug when fetching last release branch

Jonas1312:
after creating the branch release/2.130.0, the e2e CI detected the latest branch as being release/2.129.0:

image

the e2e tests "release branch against preprod" were launched on release/2.129.0, when it should've been launched on release/2.130.0

with this fix:
image

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -138,45 +138,11 @@ jobs:
         run: echo "LTS_CYCLE_FIRST_SDK_REF=release/2.129.0" >> $GITHUB_ENV # this should be updated when a new LTS cycle starts
 
       - name: Get latest release branch
-        shell: python
+        shell: shell
         run: |
-          import json
-          import os
-          import urllib.request
-          from packaging.version import InvalidVersion, parse
-
-          def parse_version(version: str):
-              try:
-                  return parse(version)
-              except InvalidVersion:
-                  return parse("0.0.0")
-
-          BRANCH_PREFIX = "release/"
-          per_page = 100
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e_tests.yml
```diff
@@ -138,13 +138,41 @@ jobs:
...
       - name: Get latest release branch
-        shell: bash
+        shell: python
         run: |
-          last_version=$(curl --silent "https://api.github.com/repos/kili-technology/kili-python-sdk/releases/latest" | jq -r .tag_name)
-          echo "Last version: $last_version"
-          IFS=. read -r major minor patch <<< "$last_version"
-          last_minor_version="$major.$minor.0"
-          echo "LASTEST_RELEASE_BRANCH=release/$last_minor_version" >> $GITHUB_ENV
+          import urllib.request
```
PierreLeveau, line 143: for long scripts like that, it can be preferrable to keep them in a separate Python file to benefit from the Python dev tooling

PierreLeveau (changes requested):

.github/workflows/e2e_tests.yml
Jonas1312: it's preferable but more prone to bugs

if we put that in a python script, the python script will have to be downloaded, since at this point in the workflow we haven't checkout the repo yet

also, we would have to parse the output of the python script and store it in a bash variable. The parsing of the python script output should be easy, but if we have a bug in the python script, it might not appear in the workflow logs, which makes debugging complicated

PierreLeveau (approved):

PR #1136, opened 2023-03-31
refactor: move code to core and entrypoints

PierreLeveau:
e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4573780239


Jonas1312 (changes requested):
many temp files in several folders

image

so you renamed the "user_interface" to "entrypoints" right?

I don't know if it's the best name. To me, I consider we have only two entrypoints: cli and kili.client 

And maybe we should keep those classes in core IMO (    MutationsAsset,    MutationsDataConnection,    MutationsIssue,    MutationsLabel,    MutationsNotification,    MutationsPlugins,    MutationsProject, etc...)

We don't want to put "services" in "core" too?

PierreLeveau:
> And maybe we should keep those classes in core IMO ( MutationsAsset, MutationsDataConnection, MutationsIssue, MutationsLabel, MutationsNotification, MutationsPlugins, MutationsProject, etc...)
I followed 
these are the client methods, so they should be part of the user interface.
that's usual to have the services in a separate folder (see https://www.cosmicpython.com/book/appendix_project_structure.html for example).
I'm removing the temp files.

Jonas1312 (changes requested):
still some *.tmp files

https://github.com/kili-technology/kili-python-sdk/tree/refactor/ml-1051-core-user-interface/tests/integration/entrypoints/queries


src\kili\services\copy_project\__init__.py.5c924411bf14d58462d07042ba76120e.tmp

kili-python-sdk\tests\integration\entrypoints\queries\test_media_downloader.py.b3169b2ecbebc9f4c5744f853e100021.tmp

maybe others

PR #1143, opened 2023-04-04
feat: add children jobs support to label parsing module

Jonas1312:
I had to create a "Project" type that contains the job interface, project input type, etc, so that we can pass just one object to all classes.

I had circular imports issues (pytest could not detect the tests). I fixed them using "import ... as ***_module".

But pylint still shows this error:
```text
************* Module kili.core.graphql.operations.organization.queries
src/kili/core/graphql/operations/organization/queries.py:1:0: R0401: Cyclic import (kili.services.label_data_parsing.job_response -> kili.services.label_data_parsing.json_response) (cyclic-import)
src/kili/core/graphql/operations/organization/queries.py:1:0: R0401: Cyclic import (kili.services.label_data_parsing.annotation -> kili.services.label_data_parsing.category -> kili.services.label_data_parsing.json_response -> kili.services.label_data_parsing.job_response) (cyclic-import)
src/kili/core/graphql/operations/organization/queries.py:1:0: R0401: Cyclic import (kili.services.label_data_parsing.annotation -> kili.services.label_data_parsing.json_response -> kili.services.label_data_parsing.job_response) (cyclic-import)
```

I think it's a pylint bug because I don't know why it's saying the errors are in module `src/kili/core/graphql/operations/organization/queries`...

Right now, all modules are at the same level in `kili.services.label_data_parsing`. The circular import comes from the fact that `json_response.py` module uses the `job_response.py` module, that then uses either (or both) the `annotation.py` module and `category.py` module. And if there are children jobs, `annotation.py` or `category.py` will import `json_response.py` to create a children job response.


Jonas1312:
I prefer not to rework the json_response.py for now. I'll split it in ChildrenParsedJobs and ParentParsedJobs later on (if needed) when I'll do the video label parsing

PierreLeveau (approved):

PR #1144, opened 2023-04-04
fix: fix graphql schema caching bug

Jonas1312:
We had this bug several times during the past few days:

```text
E       graphql.error.syntax_error.GraphQLSyntaxError: Syntax Error: Unexpected <EOF>.
E       
E       GraphQL request:1:1
E       1 |
E         | ^
```

this bug would only happen on `test_copy_project_e2e_with_ocr_metadata`

It's hard to reproduce and happens randomly.

Could be caused by writing the schema on disk and reading it just after when the writing is not finished.

Or cause by the fact that we now run tests in parallel. So we can have two kili clients caching the schema at the same time.

### DIFF
--- a/src/kili/core/graphql/graphql_client.py
+++ b/src/kili/core/graphql/graphql_client.py
@@ -33,11 +33,7 @@ class GraphQLClient:
     """GraphQL client."""
 
     def __init__(
-        self,
-        endpoint: str,
-        api_key: str,
-        client_name: GraphQLClientName,
-        verify: bool = True,
+        self, endpoint: str, api_key: str, client_name: GraphQLClientName, verify: bool = True
     ) -> None:
         self.endpoint = endpoint
         self.api_key = api_key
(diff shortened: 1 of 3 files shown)


src/kili/core/graphql/graphql_client.py
```diff
@@ -58,6 +58,10 @@ def __init__(
             method="POST",
         )
 
+        # Use mutex to avoid multiple processes operating in the cache directory at the same time
+        self._cache_dir_lock = FileLock(
```
PierreLeveau, line 63: Does this mean that the lock will be created as soon as the GraphQLCLient will be instantiated?

Jonas1312: now the lock is active when it's called with .acquire() or .__enter__

PierreLeveau (approved):

Jonas1312:
I'll rework it tomorrow, I don't like the way it is now

PR #1150, opened 2023-04-05
fix: disable label export with assets for project with cloud storage

Jonas1312:
I removed the method `_check_content_presence`, because even though the assets are on a cloud storage, the backend would still generate some signed urls, making this method useless.

In the media downloader, I raise a warning, disable media download, remove the "content"  and "jsonContent" fields 

In the fetch_assets of services.export, I crash if with_assets is enabled and we detect a data connection

e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/4625854256


src/kili/entrypoints/queries/asset/media_downloader.py
```diff
@@ -46,11 +51,31 @@ def get_download_assets_function(
...
+
+    # We need to query the data connections to know if the assets are hosted in a cloud storage
+    # If so, we remove the fields "content" and "jsonContent" from the query
+    data_connections_gen = DataConnectionsQuery(auth.client)(
+        where=DataConnectionsWhere(project_id=project_id),
+        fields=["id"],
+        options=QueryOptions(disable_tqdm=True, first=1, skip=0),
+    )
+    if len(list(data_connections_gen)) > 0:
+        warnings.warn(
```
PierreLeveau, line 63: shouldn't we rather crash ?

PierreLeveau (approved):

PR #1157, opened 2023-04-07
Add archived filter to query builder of projects

obendidi:


src/kili/entrypoints/queries/project/__init__.py
```diff
@@ -121,6 +124,8 @@ def projects(
             updated_at_gte: Returned projects should have a label whose update date is greater or equal
                 to this date.
             updated_at_lte: Returned projects should have a label whose update date is lower or equal to this date.
+            archived: If `True`, only archived projects are returned, if `False`, only active projects are returned.
+                None disable this filter.
```
PierreLeveau, line 128:
```suggestion
                `None` disables this filter.
```

PierreLeveau (changes requested):
Thanks for this PR! There is a small typo in the doc that you can fix.

PR #1158, opened 2023-04-11
build: constrain the dependencies to major versions

manuelcgallucci:
Completed requierments.txt with the libraries and the versions.


PierreLeveau (changes requested):
the dependencies are in `setup.py`. All the dependencies already listed should be constrained to the major version, like `"tenacity>=8.0.0,<9.0.0"` for example.

docs/requirements.txt
```diff
@@ -1,3 +1,31 @@
...
+charset-normalizer==3.1.0
+click==8.1.3
+cuid==0.4
+ffmpeg-python==0.2.0
+filelock==3.11.0
+future==0.18.3
+gql==3.4.0
+graphql-core==3.2.3
+idna==3.4
+kili==2.131.1
```
p-desaintchamas, line 10: Not sure this should be hardcoded, else doc will never be generated for new versions of the sdk right ?

manuelcgallucci:
Added dependencies in setup.py, only limited to mayor releases, for libraries with minor releases only, the following was defined, example:
"cuid==0.4" in pip freeze, then in setup.py -> "cuid>=0.4,<0.5"

setup.py
```diff
@@ -5,57 +5,57 @@
...
 
 install_requires = [
-    "pandas",
-    "click",
-    "requests",
-    "tabulate",
-    "tenacity",
-    "tqdm",
-    "typeguard<3.0.0",
+    "pandas>=1.0.0",
```
PierreLeveau, line 8: here and below: missing upper bound `"pandas>=1.0.0,<2.0.0"`

PierreLeveau (changes requested):

Jonas1312 (changes requested):
no need to change docs/requirements.txt

Jonas1312 (changes requested):
the docs/requirements.txt is not needed anymore?

why do we lock some packages to minor versions and not major?

the package "filelock" is gone, need to merge/rebase and fix conflicts

PierreLeveau:
> the docs/requirements.txt is not needed anymore?
> 
> why do we lock some packages to minor versions and not major?
> 
> the package "filelock" is gone, need to merge/rebase and fix conflicts

indeed, the doc requirements should be restored.
There can be minor version locking if a feature we use is only present starting from a given minor version. This is also a common practice to lock the minor when the version is <1.0

Jonas1312 (approved):

theodu (approved):

PierreLeveau (approved):

PR #1159, opened 2023-04-13
docs(data-1163): remove ignore and update docstrings for custom_clic

hguerni:


hguerni:
Capture d’écran 2023-04-13 à 11 08 26

src/kili/services/plugins/model.py
```diff
@@ -105,8 +105,21 @@ def on_custom_interface_click(
         so it should not be used.
 
         Args:
-            label: label submitted to Kili
+            label: Label submitted to Kili: a dictionary containing the following fields:
```
gheorghetutunaru, line 108: Label has only `id` and `jsonResponse`

src/kili/services/plugins/model.py
```diff
@@ -105,8 +105,21 @@ def on_custom_interface_click(
...
             asset_id: id of the asset on which the action is called
+
+        Example use:
+
+            >>> def on_custom_interface_click(self, label: Label, asset_id: str):
+            >>>     json_response = label.get('jsonResponse')
+            >>>     if label_is_respecting_business_rule(json_response):
+            >>>         return
+            >>>     else:
+            >>>         self.kili.send_back_to_queue(asset_ids=[asset_id])
```
gheorghetutunaru, line 122: No `send_back` in this example, since it should be triggered by the labeler himself. A `create_issues` would be more appropriate

src/kili/services/plugins/model.py
```diff
@@ -105,8 +105,21 @@ def on_custom_interface_click(
         so it should not be used.
```
gheorghetutunaru, line 105: I would remove the `it should not be used` part

src/kili/services/plugins/model.py
hguerni: Oh ok, thank you I was unaware

src/kili/services/plugins/model.py
hguerni: I agree, it feel weird to have the same example over and over again, and it doesn't fit this handler. Thanks for the insight

src/kili/services/plugins/model.py
```diff
@@ -102,11 +102,27 @@ def on_custom_interface_click(
...
+            >>> def on_custom_interface_click(self, label: Label, asset_id: str):
+            >>>     json_response = label.get('jsonResponse')`
+            >>>     label_id = label.get('id')
+            >>>     issue = label_is_respecting_business_rule(json_response)
+            >>>     if !issue:
+            >>>         return
+            >>>     else:
+            >>>         self.kili.create_issues(
+                            project_id=self.project_id,
+                            abel_id_array=[label],
```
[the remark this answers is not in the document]
hguerni: sorry

florianlega (approved):

PR #1160, opened 2023-04-13
fix(data-1132): replace append_to_issues by create_issues in doc and …

hguerni:
replace append_to_issues by create_issues in doc and recipes


recipes/plugins_library/plugin_document.py
```diff
@@ -41,13 +41,12 @@ def on_submit(self, label: Dict, asset_id: str) -> None:
...
         n_issues = len(text_issues_array)
 
-        for i in range(n_issues):
-            self.kili.append_to_issues(
-                label_id=label["id"],
-                project_id=project_id,
-                text=text_issues_array[i],
-                object_mid=mid_issues_array[i],
-            )
+        self.kili.create_issues(
```
gheorghetutunaru, line 44: You should add a condition on `n_issues`

recipes/plugins_library/plugin_image.py
```diff
@@ -9,13 +9,16 @@ def check_rules_on_label(label: Dict):
...
         if annotation["categories"][0]["name"] == "OBJECT_A":
             counter += 1
 
     if counter <= 1:
-        return []
-    return [f"There are too many BBox ({counter}) - Only 1 BBox of Object A accepted"]
+        return issues_array
+    return issues_array.append(
+        [f"There are too many BBox ({counter}) - Only 1 BBox of Object A accepted"]
+    )
```
gheorghetutunaru, line 21: This doesn't change anything, no ?

recipes/plugins_library/plugin_image.py
```diff
@@ -29,15 +32,14 @@ def on_submit(self, label: Dict, asset_id: str) -> None:
 
         project_id = self.project_id
 
-        if len(issues_array) > 0:
+        if issues_array and len(issues_array) > 0:
```
gheorghetutunaru, line 35: `issues_array` can be  undefined ?

recipes/plugins_library/plugin_image.py
hguerni: I had pylance type error if I didn't do that

recipes/plugins_library/plugin_document.py
hguerni: ok

recipes/plugins_library/plugin_image.py
hguerni: I don't think so but pylance does, It triggered an error, I think it is linked to the new typing of check_rule_on_label return

gheorghetutunaru: Oh, okay, wanted to understand why you changed it. 
You can also try to type the return of `check_rule_on_label` to be `List[Optional[str]]`, I think this would solve all the pylance problems

hguerni: I tried but it conflict with `[f"There are too many BBox ({counter}) - Only 1 BBox of Object A accepted"]` line 20. I will leave it as is

gheorghetutunaru: There are no errors if the `check_rule_on_label` is left unchanged (if only the type of return is added). Did you test the plugin ? Because in fact `return array.append()` returns `None`, not the array, so that's why pylint suggested that `issues_array` can be undefined : https://trinket.io/python/f7ad7f9864

image

hguerni: indeed

p-desaintchamas (approved):

gheorghetutunaru (approved):

PR #1166, opened 2023-04-20
fix: forbid export to COCO and VOC when project is connected to cloud storage

PierreLeveau:
- fix: COCO and VOC exports not possible when data connection
- refactor: isolate asset access test
- tests: fix missing content
- feat: adding specific exception for asset access
- style: pylint and cleanup
- docs: add warning

### DIFF
--- a/src/kili/entrypoints/queries/label/__init__.py
+++ b/src/kili/entrypoints/queries/label/__init__.py
@@ -689,6 +689,8 @@ class QueriesLabel:
 
         !!! warning "Cloud storage"
             Export with asset download (`with_assets=True`) is not allowed for projects connected to a cloud storage.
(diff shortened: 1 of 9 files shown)


Jonas1312 (changes requested):
it is not necessary to add a `requires_asset_access` in the children classes, since CocoExporter and VocExporter already automatically force `with_assets=True` in their `__init__` (and warn). It feels like we duplicate the logic 

In the `_check_asset_access` method, if we try a COCO export on a project not connected to a cloud storage, the export will still work because `len(list(data_connections_gen)) == 0`

We should have two separate checks:

- one that checks if the format is compatible with `with_assets=False` -> if voc and coco, we crash
- one that checks if `with_assets=True` and project connected to cloud storage -> if true, we crash

Jonas1312 (approved):

PR #1171, opened 2023-04-24
feat: add get plugin build errors

florianlega:


src/kili/entrypoints/queries/plugins/__init__.py
```diff
@@ -27,6 +27,38 @@ def __init__(self, auth: KiliAuth):
...
         self.auth = auth
 
+    @typechecked
+    def get_plugin_build_errors(
+        self,
+        plugin_name: str,
+        start_date: Optional[datetime] = None,
+        limit: Optional[int] = None,
+        skip: Optional[int] = None,
+    ):
```
Jonas1312, line 37: return type hint missing

src/kili/entrypoints/queries/plugins/__init__.py
```diff
@@ -27,6 +27,38 @@ def __init__(self, auth: KiliAuth):
...
+        """
+
+        where = PluginBuildErrorsWhere(
+            plugin_name=plugin_name, start_date=start_date
+        )
+        options = QueryOptions(
+            first=limit, skip=skip or 0, disable_tqdm=False
+        )  # disable tqm is not implemnted for this query
+        pretty_result = PluginQuery(self.auth.client).get_build_errors(where, options)
+        return json.dumps(pretty_result, sort_keys=True, indent=4)
```
Jonas1312, line 60: why convert to string?

src/kili/entrypoints/queries/plugins/__init__.py
```diff
@@ -27,6 +27,38 @@ def __init__(self, auth: KiliAuth):
...
+        skip: Optional[int] = None,
+    ):
+        # pylint: disable=line-too-long
+        """Get paginated build errors of a plugin.
+
+        Args:
+            plugin_name: Name of the plugin
+            start_date: Datetime used to get the build errors from, if not provided, it will be the plugin's creation date
+            limit: Limit for pagination, if not provided, it will be 100
+            skip: Skip for pagination, if not provided, it will be 0
```
Jonas1312, line 45: Plugin methods should behave like all other methods in the SDK: if we call the method with no `limit` or `skip`, it should return all elements (here build errors), not only the first 100 elements

Jonas1312 (changes requested):

src/kili/entrypoints/queries/plugins/__init__.py
florianlega: For indent purpose

src/kili/entrypoints/queries/plugins/__init__.py
florianlega: We can't on logs fetched from google, we need to specify a pageSize

Jonas1312: The method should automatically paginate the calls by batches of size 100, and then merge them.

Same thing as we do here for the graphql queries: https://github.com/kili-technology/kili-python-sdk/blob/91967e77596d59cc6b49dd07642b3ebefb3380fe/src/kili/core/graphql/__init__.py#L104

It should be done for the `get_plugin_logs` method too: https://python-sdk-docs.kili-technology.com/2.131/sdk/plugins/#kili.entrypoints.queries.plugins.__init__.QueriesPlugins.get_plugin_logs

The user should not see how it's done under the hood, and should not have to do all the work to fetch logs/errors by batches of 100 elements, then sort, then merge, etc.

florianlega: In the current state it is not possible, we do not know the number of pages before making the first request to Google (and moreover we do not return the index of the page). Currently on a query with a limit of 100, it takes a lot of time, so if we fetch all the logs, it will take several minutes and it's not feasible at all

src/kili/entrypoints/queries/plugins/__init__.py
```diff
@@ -27,6 +31,36 @@ def __init__(self, auth: KiliAuth):
         """
         self.auth = auth
 
+    @typechecked
+    def get_plugin_build_errors(
+        self,
+        plugin_name: str,
+        start_date: Optional[datetime] = None,
+        limit: Optional[int] = None,
```
hguerni, line 39: you said it will be with a 100 as default, why not do it here ??

src/kili/entrypoints/queries/plugins/__init__.py
```diff
@@ -2,13 +2,17 @@
 
 import json
 from datetime import datetime
-from typing import List, Optional
+from typing import Dict, List, Optional, Iterable
```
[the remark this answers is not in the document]
Jonas1312: we introduced some new pre-commit hooks yep

better rebase this branch on master actually

src/kili/entrypoints/queries/plugins/__init__.py
```diff
@@ -27,6 +31,36 @@ def __init__(self, auth: KiliAuth):
...
         self.auth = auth
 
+    @typechecked
+    def get_plugin_build_errors(
+        self,
+        plugin_name: str,
+        start_date: Optional[datetime] = None,
+        limit: Optional[int] = None,
+        skip: Optional[int] = None,
+    ) -> Iterable[Dict]:
```
hguerni, line 41: As you convert to string in return don't forget to change return value ;)

src/kili/entrypoints/queries/plugins/__init__.py
Jonas1312: wdyt @kili-technology/ml

PierreLeveau: if it is not feasible to get all the logs by default, let's keep a default number of 100 logs, and paginate if it is >100.

marc-delpech (approved):

gheorghetutunaru (approved):

PR #1183, opened 2023-04-25
fix: inverse .jobs and .frames order for video label parsing

Jonas1312:
purpose of this PR was to switch the `label.frames[42].jobs["job_name"]` to `label.jobs["job_name"].frames[42]`

done:
- removed the project input types that are not in the graphql schema anymore
- in `label.py`, the `ParsedLabel` doesn't have a `.frames` attribute anymore
- in `json_response.py`, the `ParsedJobs` inherits from two private classes `_ParsedOtherJobs` and `_ParsedVideoJobs`, and chooses the right class at runtime depending if the label is a video label or not
- I had to introduce a "fake" class `class JobPayload(FramesList, job_response_module.JobPayload):`. This class is used to expose all attributes/methods of both `FramesList` (for the `.frames` attribute), and all properties of `job_response_module.JobPayload` so that autocompletion works during coding

### DIFF
--- a/src/kili/services/label_data_parsing/json_response.py
+++ b/src/kili/services/label_data_parsing/json_response.py
@@ -25,7 +25,7 @@ class _ParsedVideoJobs:
 
         # all job names in the json response should be in the json interface too
         for _, frame_response in json_response.items():
-            for job_name in frame_response:
+            for job_name in frame_response.keys():
                 if job_name not in json_interface:
                     raise JobNotExistingError(job_name)
 
@@ -69,7 +69,7 @@ class _ParsedVideoJobs:
 
 
 # pylint: disable=too-few-public-methods
-class _ParsedOtherJobs:
+class _ParseClassicJobs:
     def __init__(
         self,
         project_info: Project,
@@ -95,18 +95,7 @@ class _ParsedOtherJobs:
             ]
 
         for job_name in job_names_to_parse:
-            job_interface = json_interface[job_name]  # type: ignore
-            job_response = json_response.get(job_name, {})  # the json response may be empty
-
-            # a required parent job should have a non-empty json response
-            if (
-                not job_interface["isChild"]  # check if parent
-                and job_interface["required"]  # check if required
-                and "VIDEO" not in project_info["inputType"]  # can have empty frames
-                and not job_response
-            ):
-                raise JobNotExistingError(job_name)
-
+            job_response = json_response.get(job_name, {})  # the json response can be empty
             self._json_data[job_name] = job_response_module.JobPayload(
                 job_name=job_name,
                 project_info=project_info,
@@ -120,7 +109,7 @@ class _ParsedOtherJobs:
         return ret
 
 
-def _is_video_job(project_info: Project, json_response: Dict) -> bool:
+def _is_video_response(project_info: Project, json_response: Dict) -> bool:
     """Returns True if the json response is a video job, False otherwise."""
     if "VIDEO" not in project_info["inputType"]:
         return False
@@ -139,7 +128,7 @@ def _is_video_job(project_info: Project, json_response: Dict) -> bool:
     return True
 
 
-class ParsedJobs(_ParsedOtherJobs, _ParsedVideoJobs):
+class ParsedJobs(_ParseClassicJobs, _ParsedVideoJobs):
     """Class for label json response parsing."""
 
     def __init__(
@@ -159,9 +148,9 @@ class ParsedJobs(_ParsedOtherJobs, _ParsedVideoJobs):
             json_response: Value of the key "jsonResponse" of a label.
             job_names_to_parse: List of job names to parse. By default, parse all the jobs that are not children.
         """
-        self._is_video_job = _is_video_job(project_info, json_response)
+        self._is_video_response = _is_video_response(project_info, json_response)
 
-        if self._is_video_job:
+        if self._is_video_response:
             _ParsedVideoJobs.__init__(
                 self,
                 project_info=project_info,
@@ -169,7 +158,7 @@ class ParsedJobs(_ParsedOtherJobs, _ParsedVideoJobs):
                 job_names_to_parse=job_names_to_parse,
             )
         else:
-            _ParsedOtherJobs.__init__(
+            _ParseClassicJobs.__init__(
                 self,
                 project_info=project_info,
                 json_response=json_response,
@@ -178,9 +167,9 @@ class ParsedJobs(_ParsedOtherJobs, _ParsedVideoJobs):
 
     def to_dict(self) -> Dict[str, Dict]:
         """Returns the parsed json response as a dict."""
-        if self._is_video_job:
+        if self._is_video_response:
             return _ParsedVideoJobs.to_dict(self)
-        return _ParsedOtherJobs.to_dict(self)
+        return _ParseClassicJobs.to_dict(self)
 
     def __repr__(self) -> str:
         """Returns the representation of the object."""
--- a/src/kili/services/label_data_parsing/json_response.py
+++ b/src/kili/services/label_data_parsing/json_response.py
@@ -165,7 +165,9 @@ class ParsedVideoJobs(_ParsedJobs):
                         project_info=project_info,
                         job_payload=job_response,
                     )
-                    for _, frame_json_response in json_response.items()
+                    for _, frame_json_response in sorted(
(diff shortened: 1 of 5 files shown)


src/kili/services/label_data_parsing/json_response.py
```diff
@@ -66,23 +113,75 @@ def __init__(
...
+
+        This class will modify the input json_response.
+        If you want to keep the original json_response, use deepcopy.
+
+        Args:
+            project_info: Information about the project.
+            json_response: Value of the key "jsonResponse" of a label.
+            job_names_to_parse: List of job names to parse. By default, parse all the jobs that are not children.
+        """
+        self._is_video_job = _is_video_job(project_info, json_response)
```
theodu, line 162: is_video_response

src/kili/services/label_data_parsing/json_response.py
```diff
@@ -66,23 +113,75 @@ def __init__(
...
+
+    if not all(frame_id.isdigit() for frame_id in json_response.keys()):
+        return False
+
+    for frame_response in json_response.values():
+        if len(frame_response) == 0:
+            continue
+
+        for job_name in frame_response.keys():
+            if not (isinstance(job_name, str) and job_name in project_info["jsonInterface"]):
```
theodu, line 136: why not to return a JobNotExistingError(job_name) here as in other jobs ? ?

src/kili/services/label_data_parsing/json_response.py
```diff
@@ -66,23 +113,75 @@ def __init__(
...
         return ret
 
+
+def _is_video_job(project_info: Project, json_response: Dict) -> bool:
+    """Returns True if the json response is a video job, False otherwise."""
+    if "VIDEO" not in project_info["inputType"]:
+        return False
+
+    if not all(frame_id.isdigit() for frame_id in json_response.keys()):
+        return False
```
theodu, line 129: why not to return an error rather to try to parse it with the ParseOtherJobs class ?
and a more general comment: isn't the test with the inputType enough to know if it is a video response or not ?

src/kili/services/label_data_parsing/json_response.py
```diff
@@ -66,23 +113,75 @@ def __init__(
...
-            raise JobNotExistingError(job_name)
-        return self._json_data[job_name]
-
-    @property
-    def jobs(self) -> "ParsedJobs":
-        """Returns self."""
-        return self
-
-    def to_dict(self) -> Dict:
+    def to_dict(self) -> Dict[str, Dict]:
```
theodu, line 116: comment for above that was in a previous PR:
in ParseOtherJobs, you test:
```
# a required parent job should have a non-empty json response
            if (
                not job_interface["isChild"]  # check if parent
                and job_interface["required"]  # check if required
                and "VIDEO" not in project_info["inputType"]  # can have empty frames
                and not job_response
            ):
                raise JobNotExistingError(job_name)
```
But we agreed with PM when I was working on json response validation that one can import a label from the API/SDK with empty response for required jobs. It is very useful for predictions labels for example when you want to upload predictions only for some jobs. But in the general case, the required feature only work when labeling on the platform.
It should be changed. Can be in another PR

src/kili/services/label_data_parsing/json_response.py
```diff
@@ -1,34 +1,81 @@
...
+class _ParsedOtherJobs:
+    def __init__(
+        self,
+        project_info: Project,
+        json_response: Dict,
+        job_names_to_parse: Optional[List[str]] = None,
+    ) -> None:
         self._json_data: Dict[str, "job_response_module.JobPayload"] = {}
 
         json_interface = project_info["jsonInterface"]
```
theodu, line 81: comment for a previous PR
in ParseOtherJobs, you test:
```
# a required parent job should have a non-empty json response
            if (
                not job_interface["isChild"]  # check if parent
                and job_interface["required"]  # check if required
                and "VIDEO" not in project_info["inputType"]  # can have empty frames
                and not job_response
            ):
                raise JobNotExistingError(job_name)
```
But we agreed with PM when I was working on json response validation that one can import a label from the API/SDK with empty response for required jobs. It is very useful for predictions labels for example when you want to upload predictions only for some jobs. But in the general case, the required feature only work when labeling on the platform.
It should be changed. Can be in another PR

src/kili/services/label_data_parsing/json_response.py
theodu: I see that you raise the error inside ParsedVideoJobs. 
According to me, only the input type should decide if we parse it the video way of the classic way

src/kili/services/label_data_parsing/json_response.py
```diff
@@ -1,34 +1,81 @@
...
+        for job_name, frames_list in self._json_data.items():
+            for frame_id, job_payload in enumerate(frames_list):
+                job_payload_dict = job_payload.to_dict()
+                if job_payload_dict:
+                    ret[str(frame_id)][job_name] = job_payload_dict
+        return ret
+
+
+# pylint: disable=too-few-public-methods
+class _ParsedOtherJobs:
```
theodu, line 72: _ParseNonVideoJobs ? or _ParseClassicJobs ?
The "other" is a bad name pattern I think

src/kili/services/label_data_parsing/json_response.py
```diff
@@ -99,14 +198,55 @@ def __iter__(self) -> Iterator[str]:
...
+            raise FrameIndexError(frame_index=key, nb_frames=len(self))
+        return super().__getitem__(key)
+
+    @property
+    def frames(self) -> "FramesList":
+        """Returns the list of frames."""
+        return self
+
+
+class JobPayload(FramesList, job_response_module.JobPayload):
```
theodu, line 240: are you able to keep good type hints with this ?
It will always propose you the attibutes frames even if it is not a video project isn't it ?
(I don't see how to do it another way)

theodu (changes requested):
.

src/kili/services/label_data_parsing/json_response.py
PierreLeveau: both tests are relevant, maybe the second one could be in another method to separate the concerns

src/kili/services/label_data_parsing/json_response.py
PierreLeveau: since it both tests the project_info and json_response, the name is fine

src/kili/services/label_data_parsing/json_response.py
PierreLeveau: or simply _ParsedJobs (not specific)

src/kili/services/label_data_parsing/json_response.py
Jonas1312: this method is just a helper to determine if it's a video json response or not

src/kili/services/label_data_parsing/json_response.py
Jonas1312: No the input type is not enough

The ParsedJobs class can be use several times for the same (root) json response

For example, for a video project, the root json response will be parsed using _ParsedVideoJobs

but if the json response contains children jobs, those job json responses will need to be parsed using _ParsedOtherJobs, despite the input_type being "VIDEO"

src/kili/services/label_data_parsing/json_response.py
PierreLeveau: we have all the possible fields (annotations, etc.), and this can only be decided at runtime, lgtm

PierreLeveau (approved):

src/kili/services/label_data_parsing/json_response.py
Jonas1312: type hints work yes, otherwise pyright CI would fail

Well during run time, the program will crash if you try to access `label.jobs["job_name"].frames` on a non-video label yes

But since at static time, we do not know the label input type, we still have to expose the attribute `.frames`, for autocompletion, hover peeks, etc

and yes that's the only way I found to do it

src/kili/services/label_data_parsing/json_response.py
Jonas1312: oh ok I didn't know that, thanks

src/kili/services/label_data_parsing/json_response.py
theodu: Ah yeah I get it, thanks

theodu (approved):

PR #1185, opened 2023-04-26
perf: improve Kili client initialization time

Jonas1312:
done:
- stop storing user id and email in kiliauth saves 100ms
- add an env var `KILI_SDK_SKIP_CHECKS` that disables checks and graphql schema related stuff

`from kili.client import Kili` takes around ~700ms (mostly because of pandas)

`kili = Kili()` takes ~0s when checks are disabled

to consider:
- pandas import takes ~500ms...: It's pretty easy to add lazy loading of import. But I haven't found yet how to type pd.Dataframe 🤔  We can keep it as is for now. Maybe updating to more recent versions of pandas could improve this (we are using old version now because of python3.7)

### DIFF
--- a/src/kili/core/authentication.py
+++ b/src/kili/core/authentication.py
@@ -1,4 +1,5 @@
 """API authentication module."""
+import os
 import warnings
 from datetime import datetime, timedelta
 from typing import Dict
@@ -23,12 +24,23 @@ class KiliAuth:
     def __init__(
         self, api_key: str, api_endpoint: str, client_name: GraphQLClientName, verify=True
     ) -> None:
+        """Initialize KiliAuth.
+
+        Args:
+            api_key: Kili API key.
+            api_endpoint: Kili API endpoint.
+            client_name: Name of the client.
+            verify: Whether to verify the SSL certificate.
+        """
         self.api_key = api_key
         self.api_endpoint = api_endpoint
         self.client_name = client_name
         self.verify = verify
 
-        self.check_api_key_valid()
+        skip_checks = os.environ.get("KILI_SDK_SKIP_CHECKS", None) is not None
+
+        if not skip_checks:
+            self.check_api_key_valid()
 
         self.client = GraphQLClient(
             endpoint=api_endpoint,
@@ -37,11 +49,8 @@ class KiliAuth:
             verify=self.verify,
         )
 
-        self.check_expiry_of_key_is_close()
-
-        user = self.get_user()
-        self.user_id = user["id"]
-        self.user_email = user["email"]
+        if not skip_checks:
+            self.check_expiry_of_key_is_close()
 
     def check_api_key_valid(self) -> None:
         """Check that the api_key provided is valid."""
--- a/tests/e2e/test_client.py
+++ b/tests/e2e/test_client.py
@@ -22,14 +22,3 @@ def test_client_init_not_too_long_with_checks_disabled():
     time_spent = timer.timeit(number=1)
 
     assert time_spent < 0.1
-
-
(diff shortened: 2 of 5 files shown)


src/kili/core/authentication.py
```diff
@@ -23,12 +24,23 @@ class KiliAuth:
...
+            client_name: Name of the client.
+            verify: Whether to verify the SSL certificate.
+        """
         self.api_key = api_key
         self.api_endpoint = api_endpoint
         self.client_name = client_name
         self.verify = verify
 
-        self.check_api_key_valid()
+        skip_checks = os.environ.get("KILI_SDK_SKIP_CHECKS", None) is not None
```
theodu, line 40: maybe that data quality can set this environment variable in the env that runs the plugin in google ?

tests/e2e/test_client.py
```diff
@@ -0,0 +1,35 @@
...
+
+def test_client_with_checks_disabled_always_faster_than_with_checks_enabled():
+    timer_with_checks = timeit.Timer("_ = Kili()", setup="from kili.client import Kili")
+    time_spent_with_checks = timer_with_checks.timeit(number=1)
+
+    with mock.patch.dict("os.environ", {"KILI_SDK_SKIP_CHECKS": "True"}):
+        timer_without_checks = timeit.Timer("_ = Kili()", setup="from kili.client import Kili")
+        time_spent_without_checks = timer_without_checks.timeit(number=1)
+
+    assert time_spent_with_checks > 10 * time_spent_without_checks  # one order of magnitude
```
theodu, line 35: if we decrease by a lot the time spent to init the clients with checks, we don't want this test to fail.
We can remove it in my opinion

theodu (changes requested):
for pandas, can't you do like this ?
```
def label_as_df:
    import pandas as pd
    ...
    return pd.dataFrame(labels)
```

gheorghetutunaru (approved):

src/kili/core/authentication.py
Jonas1312: yep! It should improve the init delay

theodu: @p-desaintchamas @gheorghetutunaru

theodu:
> > for pandas, can't you do like this ?
> > ```
> > def label_as_df:
> >     import pandas as pd
> >     ...
> >     return pd.dataFrame(labels)
> > ```
> 
> yes that's the lazy importing I was talking about. But how do I add return type hints to kili.assets()? `-> Union[Iterable[Dict], pd.DataFrame]:`

Ah yeah I get it. first, we should add to do not support pandas anymore in the next major release.
I tried to remove the return type of the last assets overload method and it worked. Vscode still infer generator or list depending of the as_generator argument but we don't use pd.dataFrame anymore before to import it

Jonas1312:
> > > for pandas, can't you do like this ?
> > > ```
> > > def label_as_df:
> > >     import pandas as pd
> > >     ...
> > >     return pd.dataFrame(labels)
> > > ```
> > 
> > 
> > yes that's the lazy importing I was talking about. But how do I add return type hints to kili.assets()? `-> Union[Iterable[Dict], pd.DataFrame]:`
> 
> Ah yeah I get it. first, we should add to do not support pandas anymore in the next major release. I tried to remove the return type of the last assets overload method and it worked. Vscode still infer generator or list depending of the as_generator argument but we don't use pd.dataFrame anymore before to import it

Not sure to understand what you did. You removed the return type hints of the kili.assets method or?

Push a commit on this branch with your fix if you want

PierreLeveau (approved):

theodu (approved):

PR #1186, opened 2023-04-26
feat: add warning message when assets are filtered because of duplicate external ids

theodu:
Capture d’écran 2023-04-26 à 16 53 05

### DIFF
--- a/src/kili/services/asset_import/base.py
+++ b/src/kili/services/asset_import/base.py
@@ -428,11 +428,11 @@ class BaseAssetImporter:
             raise ImportValidationError(
                 "No assets to import, all given external_ids already exist in the project"
             )
-        nb_filtered_assets = len(assets) - len(filtered_assets)
-        if nb_filtered_assets > 0:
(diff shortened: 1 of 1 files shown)


src/kili/services/asset_import/base.py
```diff
@@ -427,6 +428,15 @@ def filter_duplicate_external_ids(self, assets):
             raise ImportValidationError(
                 "No assets to import, all given external_ids already exist in the project"
             )
+        nb_filtered_assets = len(assets) - len(filtered_assets)
```
hugo-mailfait-kili, line 431: Isn't it simpler to go with `len(assets_in_project)`?

theodu: I think that I used a miunderstanding name, it is nb_duplicate_assets

theodu:
> li/services/asset_import/base.py

I thought of how to do it and there is no simple hack because we get this information before to show the progress bar and to upload assets.
Does it directly come from a client?

PierreLeveau (approved):

PR #1189, opened 2023-04-27
docs: add warning for azure blobs with credentials

Jonas1312:


PierreLeveau (approved):

src/kili/entrypoints/mutations/data_connection/__init__.py
```diff
@@ -99,6 +100,11 @@ def synchronize_cloud_storage_connection(
 
         Returns:
             A dict with the DataConnection ID.
+
+        !!! warning "Azure Blob Storage"
+            This method currently does not work for Azure cloud storage connection using credentials mode.
+            Use service account mode instead or use the Kili app to synchronize the data connection.
```
theodu, line 106: what is the service account mode ?

Jonas1312: https://docs.kili-technology.com/docs/creating-an-azure-blob-storage-integration#add-service-account

PierreLeveau: Maybe we can add this link in the docstring?

PierreLeveau: ah ah missed that

PR #1282, opened 2023-06-22
fix: only return mutated user in append_to_roles 

Jonas1312:


src/kili/entrypoints/mutations/project/__init__.py
```diff
@@ -61,8 +65,7 @@ def append_to_roles(
             role: One of {"ADMIN", "TEAM_MANAGER", "REVIEWER", "LABELER"}.
```
PierreLeveau, line 65: is this line redundant given the added argument type?

theodu (approved):

PierreLeveau (approved):

PR #1283, opened 2023-06-22
fix: add retries when gql gets 401 server errors

Jonas1312:
e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/5347024792

### DIFF
--- a/src/kili/core/graphql/graphql_client.py
+++ b/src/kili/core/graphql/graphql_client.py
@@ -22,6 +22,10 @@ from gql.transport.exceptions import TransportServerError
 from gql.transport.requests import RequestsHTTPTransport
 from gql.transport.requests import log as gql_requests_logger
 from graphql import DocumentNode, print_schema
+from tenacity import Retrying
+from tenacity.retry import retry_if_exception_cause_type
+from tenacity.stop import stop_after_delay
+from tenacity.wait import wait_exponential
 from typing_extensions import LiteralString
 
 from kili import __version__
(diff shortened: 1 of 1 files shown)


src/kili/core/graphql/graphql_client.py
```diff
@@ -238,38 +242,43 @@ def _execute(
                     },
                     **kwargs,
                 )
-            except (exceptions.TransportQueryError, graphql.GraphQLError) as err:
+            except (
+                graphql.GraphQLError,
+                exceptions.TransportError,
+                exceptions.TransportQueryError,
+            ) as err:
```
PierreLeveau, line 249: can be:
```
except graphql.GraphQLError as err:
    raise GraphQLError(error=err.message) from err
except exceptions.TransportQueryError as err:
    raise raise GraphQLError(error=err.errors) from err
```
...

src/kili/core/graphql/graphql_client.py
```diff
@@ -238,38 +242,43 @@ def _execute(
...
+            return _execute(document, variables, **kwargs)
+        except GraphQLError as err:
+            # if error is due do parsing or local validation of the query (graphql.GraphQLError)
+            # we refresh the schema and retry once
+            if isinstance(err.__cause__, graphql.GraphQLError):
+                self._purge_graphql_schema_cache_dir()
+                self._gql_client = self._initizalize_graphql_client()
+                return _execute(document, variables, **kwargs)  # if it fails again, we crash here
+
+            # if backend returned a 401 error, we retry a few times
```
PierreLeveau, line 269: if the api key is wrong, this means that the user will have to wait 30 seconds before being notified?

PierreLeveau (changes requested):

src/kili/core/graphql/graphql_client.py
Jonas1312: if the api key is wrong, it'll be caught during the Kili() init in Kiliauth

PierreLeveau (approved):

PR #1300, opened 2023-06-29
ci: use pypi trusted publisher

Jonas1312:
I don't have access to our pypi account

@PierreLeveau you have to follow this tutorial: https://docs.pypi.org/trusted-publishers/adding-a-publisher/

on test pypi account too

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -7,6 +7,8 @@ jobs:
   publish:
     name: Publish
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -35,13 +37,14 @@ jobs:
...
-        run: |
-          python -m twine upload --repository-url https://test.pypi.org/legacy/ dist/*
-          python -m twine upload dist/*
+      - name: Publish to test PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository-url: https://test.pypi.org/legacy/
+
+      - name: Publish to PyPI
+        if: startsWith(github.ref, 'refs/tags')
```
PierreLeveau, line 46: why is this added?

Jonas1312: to check that the commit has a tag: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#publishing-the-distribution-to-pypi-and-testpypi

PR #1308, opened 2023-07-03
fix: fix asset mutation when page_resolutions_array is None

Jonas1312:


src/kili/entrypoints/mutations/asset/helpers.py
```diff
@@ -38,9 +38,10 @@ def process_update_properties_in_assets_parameters(properties) -> Dict:
         map(is_none_or_empty, properties["to_be_labeled_by_array"])
     )
 
-    properties["page_resolutions_array"] = _ensure_page_resolution_dicts(
-        properties["page_resolutions_array"]
-    )
+    if "page_resolutions_array" in properties:
```
PierreLeveau, line 41: "page_resolutions_array" is always in properties

PierreLeveau (changes requested):

src/kili/entrypoints/mutations/asset/helpers.py
Jonas1312: not for `change_asset_external_ids`

PierreLeveau: indeed

PierreLeveau: merging the new branch that will make this change not necessary

PR #1330, opened 2023-07-12
feat: geojson

Jonas1312:


src/kili/utils/labels/geojson/collection.py
```diff
@@ -0,0 +1,211 @@
...
+        warnings.warn(f"Jobs {jobs_skipped} cannot be exported to GeoJson format.", stacklevel=2)
+    if ann_tools_not_supported:
+        warnings.warn(
+            f"Annotation tools {ann_tools_not_supported} are not supported and will be skipped.",
+            stacklevel=2,
+        )
+    return features_to_feature_collection(features)
+
+
+def geojson_feature_collection_to_json_response(
```
PierreLeveau, line 163: `geojson_feature_collection_to_kili_json_response`

src/kili/utils/labels/geojson/line.py
```diff
@@ -0,0 +1,135 @@
+"""Geojson linestring utilities."""
+from typing import Any, Dict, List, Literal, Optional, Union
+
+
+def kili_line_to_geojson_linestring(
+    polyline: List[Dict[str, float]]
+) -> Dict[Literal["type", "coordinates"], Union[Literal["LineString"], List[List[float]]]]:
```
PierreLeveau, line 7: defining a TypeDict for the return type would be cleaner?

src/kili/utils/labels/geojson/bbox.py
```diff
@@ -0,0 +1,208 @@
...
+    if job_name is not None:
+        ret["properties"]["kili"]["job"] = job_name
+    return ret
+
+
+def geojson_polygon_feature_to_kili_bbox_annotation(polygon: Dict[str, Any]) -> Dict[str, Any]:
+    """Convert a geojson polygon feature to a Kili bounding box annotation.
+
+    Args:
+        polygon: a geojson polygon feature.
```
PierreLeveau, line 135: not true, this requires custom fields in `properties`. We should choose if we require these fields or not. If we do, this should be clearly stated in the doc. IMHO we should require it, even if this limits the usability of the function.

This comment applies to all the geojson_... methods.

PierreLeveau (changes requested):

src/kili/utils/labels/geojson/point.py
```diff
@@ -0,0 +1,123 @@
...
+        {
+            'type': 'Feature',
+            'geometry': {
+                'type': 'Point',
+                'coordinates': [-79.0, -3.0]},
+                'id': 'mid_object',
+                'properties': {
+                    'kili': {
+                        'categories': [{'name': 'A'}],
+                        'children': {},
```
theodu, line 60: what is the point of writing children too ?

tests/services/label_data_parsing/mutation/test_mutation_exceptions.py
```diff
@@ -324,52 +324,6 @@ def test_add_annotation_ner_wrong_category_name():
...
-                }
-            ]
-        }
-    }
-
-    project_info = Project(jsonInterface=json_interface["jobs"], inputType="IMAGE")  # type: ignore
-    parsed_jobs = ParsedJobs(json_response=json_resp, project_info=project_info)
-    with pytest.raises(ValueError):
-        parsed_jobs["JOB_0"].annotations[0].bounding_poly[0].normalized_vertices = [
-            {"x": 10000, "y": 0}
```
theodu, line 369: for me, this should still be forbidden since it is not normalized isn't it ?

tests/utils/labels/test_geojson.py
```diff
@@ -0,0 +1,533 @@
...
+        },
+    }
+
+    # geojson to kili
+    output = geojson_point_feature_to_kili_point_annotation(geojson_point_feat)
+    assert output == kili_point_annotation
+
+
+def test_kili_bbox_to_geojson_polygon():
+    normalized_vertices = [
```
theodu, line 68: these are not normalized. we should rather introduce a new field `vertices`

src/kili/services/label_data_parsing/bounding_poly.py
```diff
@@ -76,10 +76,6 @@ def normalized_vertices(
         vertices = normalized_vertices if self._type_of_tool else normalized_vertices[0]
         for vertex in vertices:
             assert isinstance(vertex, dict), f"Vertex {vertex} is not a dict."
-            if not 0 <= vertex["x"] <= 1:
```
theodu, line 79: for me, this should stay here add we should add a new field for not normalized vertices

theodu (changes requested):
The tutorial and the doc is of very high quality. Good job!!
Just some one main remark regarding the use of `normalized_vertices`.

src/kili/utils/labels/geojson/point.py
Jonas1312: I just copy what's in the kili annotation

src/kili/utils/labels/geojson/bbox.py
Jonas1312: I modified the docstrings to explain that the kili property must be set with the categories, etc.

PierreLeveau (approved):

PR #1342, opened 2023-07-26
feat: `kili.import_text_from_csv()`

Jonas1312:


src/kili/services/import_csv/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+                "Cannot create the project if categories are not provided. You must provide either"
+                " `unique_categories` or `category_column`."
+            )
+
+        project_id = _create_project(kili, project_title, project_description, categories)
+
+        logger.info(f"Created classification project {project_title} with id {project_id}")
+
+    logger.info(f"Adding {len(content_array)} assets to project {project_id}")
+    kili.append_many_to_dataset(
```
PierreLeveau, line 55: we do not call an entrypoint in a service (we cannot call a service either)

src/kili/services/import_csv/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+
+    if category_column is None:
+        return project_id
+
+    assert categories_array is not None, categories_array
+
+    logger.info("Adding labels to project.")
+    json_response_array = _get_json_response_array(categories_array)
+    kili.append_labels(
```
PierreLeveau, line 69: same as above

src/kili/services/import_csv/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+            "CLASSIFICATION_JOB": {
+                "content": {"categories": categories_dict, "input": "radio"},
+                "instruction": "Classification job",
+                "mlTask": "CLASSIFICATION",
+                "required": 0,
+                "isChild": False,
+            }
+        }
+    }
+    project_id = kili.create_project(
```
PierreLeveau, line 123: same as above

PierreLeveau (changes requested):

src/kili/services/import_csv/__init__.py
Jonas1312: if we can't call either entrypoint nor service code, aren't we just going to duplicate code in the end? not sure if it's worth it

src/kili/entrypoints/mutations/project/import_csv.py
```diff
@@ -0,0 +1,141 @@
...
+    content_array, categories_array, external_id_array = _read_csv(
+        csv_file, content_column, category_column, external_id_column, csv_separator
+    )
+
+    if project_id is None:
+        if project_title is None:
+            raise ValueError("`project_title` is required if `project_id` is not provided.")
+
+        if project_description is None:
+            raise ValueError("`project_description` is required if `project_id` is not provided.")
```
theodu, line 38: the project description is not a required argument in the `create_project` function

src/kili/entrypoints/mutations/project/__init__.py
```diff
@@ -444,3 +446,58 @@ def update_project_anonymization(
...
+            label_type: Type of the labels to create. Only used if `category_column` is provided.
+            csv_separator: Separator used in the csv file.
+            project_id: Identifier of the project. If None, a text project is created.
+            project_title: Title of the project. Only required if `project_id` is None.
+            project_description: Description of the project. Only required if `project_id` is None.
+
+        Returns:
+            A dictionary with the project `id`.
+        """
+        project_id = import_text_assets_labels_from_csv(
```
theodu, line 490: product side question:
I really like the fact to be able to import labels in the same time. We wanted to add this feature in append_many_to_dataset in the SDK v3.
I never thought of the possibility to create a project in the same time, I find it also interesting to be able to create a simple classification project if not project_id is provided.
My only problem is that it participates in the SDK heterogeneity, and thus onboarding issues for someone discovering the SDK. 
We have only one use case where we have a function for project creation, asset import and label import and not others. 

Plus the fact that the function does 3 things make it hard to take in hands (nothing very though, I think that you handled it very well but it still have some complexity)

Why didn't we added a parameter `from_csv` in append_many_to_dataset and crashing an error if the project is not a text project ?
You maybe already had the discussion the Pierre of the product team but I am curious to know where this idea of handling everything with this function come from

src/kili/entrypoints/mutations/project/import_csv.py
```diff
@@ -0,0 +1,141 @@
...
+
+        if unique_categories:
+            categories = unique_categories
+        elif categories_array:
+            categories = list(set(categories_array))
```
theodu, line 43: we could raise an error also if categories_array is not included in unique_categories

src/kili/entrypoints/mutations/project/import_csv.py
```diff
@@ -0,0 +1,141 @@
...
+    external_id_array: List[str] = []
+
+    with open(csv_file, newline="", encoding="utf-8") as file:
+        reader = csv.DictReader(file, delimiter=csv_separator)
+
+        for i, row in enumerate(reader):
+            content_array.append(row[content_column])
+
+            if category_column is not None and categories_array is not None:
+                categories_array.append(row[category_column])
```
theodu, line 97: if one row does not have any label (ie no category given), I guess that it will insert `''` in the category array this will create issues while appending labels

theodu (changes requested):

Jonas1312:
Benjamin agreed with you @theodu , best keep it simple but include it in append_many_to_dataset 👍

src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -88,6 +92,16 @@ def append_many_to_dataset(
...
                 If `False`, the function will return faster but the assets might not be fully processed by the server.
+            from_csv: Path to a csv file containing the text assets to import.
+                Only used for `TEXT` projects. If provided, `content_array` must not be provided.
+            csv_separator: Separator used in the csv file. Only used if `from_csv` is provided.
+            csv_content_column: Name of the column containing the text assets to import.
+                Only used if `from_csv` is provided.
+                If None, the first column of the csv file will be used.
+            csv_external_id_column: Name of the column containing the external ids of the assets to import.
+                Only used if `from_csv` is provided.
+                If None, the external ids will be set to the passed `external_id_array` or to the csv row index.
```
theodu, line 103: in order to avoid complexity, I think that we should force the user to give specific column name. On the app, we can import assets from url given in a csv and the column name should explicitely be `url` and `externalId' (https://docs.kili-technology.com/docs/adding-assets-to-project#required-csv-file-format)
I think that this is giving too much freedom for something for column names that will always be the same for the user and that will increase the complexity of the function and the code

theodu: what do you think ?

Jonas1312: yep I agree

theodu: I was thinking about removing completely `csv_external_id_column` and `csv_content_column`and to rather always search for the column named `content` and `externalId`in the csv

src/kili/services/asset_import_csv/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def get_text_assets_from_csv(from_csv: str, csv_separator: str) -> Tuple[List[str], List[str]]:
+    """Get text assets from a csv file."""
+    content_array: List[str] = []
+    external_id_array: List[str] = []
+
+    with open(from_csv, newline="", encoding="utf-8") as file:
+        reader = csv.DictReader(file, delimiter=csv_separator)
+
```
theodu, line 13: we should raise an error if the csv is not well formatted

src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -103,6 +110,15 @@ def append_many_to_dataset(
             - For more detailed examples on how to import text assets,
                 see [the recipe](https://python-sdk-docs.kili-technology.com/latest/sdk/tutorials/import_text_assets/).
         """
+        if from_csv is not None:
+            if content_array is not None:
+                raise ValueError(
+                    "You cannot provide both `content_array` and `from_csv` arguments."
+                )
```
theodu, line 117: another value error id external_id_array is given too ?

theodu (changes requested):

src/kili/services/asset_import_csv/__init__.py
Jonas1312: csv reader will already throw a KeyError: 'content' or KeyError: 'externalId'

theodu (approved):

PR #1345, opened 2023-07-26
feat: implement GeoJson exporter for kili.export_labels

Jonas1312:


theodu (approved):

src/kili/services/export/tools.py
```diff
@@ -2,36 +2,43 @@
 import warnings
 from typing import Dict, List, Optional
 
+import requests
```
PierreLeveau, line 5: Since the last merge on main, we do not use requests directly, but rather `http_client` which is instantiated in the `kili.client`. It can be the opportunity to remove the dependency of the export service to kili, and make it depend on http_client and graphql_client instead.

src/kili/services/export/tools.py
```diff
@@ -164,3 +171,25 @@ def get_fields_to_fetch(export_type: ExportType):
     if export_type == "latest":
         return LATEST_LABEL_FIELDS
     return DEFAULT_FIELDS
+
+
+def is_geotiff_asset_with_lat_lon_coords(asset: Dict) -> bool:
```
PierreLeveau, line 176: `, http_client: requests.Session)`

src/kili/services/export/tools.py
```diff
@@ -164,3 +171,25 @@ def get_fields_to_fetch(export_type: ExportType):
...
+    """Check if asset is a geotiff with lat/lon coordinates."""
+    if "jsonContent" not in asset:
+        return False
+
+    if isinstance(asset["jsonContent"], str) and asset["jsonContent"].startswith("http"):
+        response = requests.get(asset["jsonContent"], timeout=30)
```
PierreLeveau, line 182: `http_client.get(...`

tests/services/export/test_export.py
```diff
@@ -535,15 +535,75 @@ def get_file_tree(folder: str):
...
 )
-@patch.object(ProjectQuery, "__call__", side_effect=mocked_ProjectQuery)
-@patch.object(AssetQuery, "__call__", side_effect=mocked_AssetQuery)
-@patch.object(AssetQuery, "count", side_effect=mocked_AssetQuery_count)
-@patch("kili.services.export.media.video.ffmpeg")
-def test_export_service_layout(
-    mocker_ffmpeg, mocker_asset_count, mocker_asset, mocker_project, name, test_case
-):
+def test_export_service_layout(mocker: pytest_mock.MockerFixture, name, test_case):
+    mocker.patch.object(AssetQuery, "count", side_effect=mocked_AssetQuery_count)
```
PierreLeveau, line 598: 👍

PierreLeveau (changes requested):
👍 but some changes necessary to deal with the new way to do requests

src/kili/services/export/tools.py
Jonas1312: I started doing it but it introduces a lot of changes for this PR, we should cover that during the SDK refactoring

PierreLeveau: ok 👍  or in a different PR. That can be quick to do.

PR #1353, opened 2023-07-27
feat: add `kili.delete_labels()`

Jonas1312:
to merge when the MR on kili staging is merged

### DIFF
--- a/src/kili/entrypoints/mutations/label/__init__.py
+++ b/src/kili/entrypoints/mutations/label/__init__.py
@@ -329,8 +329,10 @@ class MutationsLabel(BaseOperationEntrypointMixin):
         return self.format_result("data", result, Label)
 
     @typechecked
-    def delete_labels(self, ids: List[str]):
+    def delete_labels(self, ids: List[str]) -> List[str]:
         """Delete labels.
+
+        Currently, only `PREDICTION` and `INFERENCE` labels can be deleted.
 
         Args:
             ids: List of label ids to delete.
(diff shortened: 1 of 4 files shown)


src/kili/entrypoints/mutations/label/__init__.py
```diff
@@ -325,3 +326,17 @@ def create_honeypot(
...
+    def delete_labels(self, ids: List[str]):
+        """Delete labels.
+
+        Args:
+            ids: List of label ids to delete.
+
+        Returns:
+            The deleted label ids.
+        """
+        variables = {"ids": ids}
```
PierreLeveau, line 340: we should batch the call for a number of labels > 100

PierreLeveau (changes requested):

theodu (approved):

tests/integration/entrypoints/mutations/test_labels.py
```diff
@@ -0,0 +1,16 @@
...
+from kili.entrypoints.mutations.label import GQL_DELETE_LABELS, MutationsLabel
+
+
+def test_delete_labels(mocker: pytest_mock.MockerFixture):
+    kili = MutationsLabel()
+    kili.graphql_client = mocker.MagicMock()
+    kili.http_client = mocker.MagicMock()
+
+    kili.delete_labels(ids=["id1", "id2"])
+
```
PierreLeveau, line 12: i would put 101 labels and check that the client has been called twice

PierreLeveau (changes requested):

theodu:
```kili.delete_labels(ids=[])

GraphQLError: GraphQL error at index 0: WHERE parameter "id" has invalid "undefined" value"
```

we might need to do an early return if we have an empty list

theodu (changes requested):
```kili.delete_labels(ids=[])
GraphQLError: GraphQL error at index 0: WHERE parameter "id" has invalid "undefined" value"
```

we might need to do an early return if we have an empty list

theodu (approved):

PR #1356, opened 2023-07-27
feat: return asset ids in new resolver appendManyAssets

theodu:


src/kili/services/asset_import/base.py
```diff
@@ -459,9 +468,10 @@ def import_assets_by_batch(
...
 
-        responses = []
+        created_asset_ids = []
         for i, batch_assets in enumerate(batch_generator):
             # check last batch only
             verify = i == (len(batch_generator) - 1) and self.verify
-            responses.append(batch_importer.import_batch(batch_assets, verify))
-        return responses[-1]
+            batch_created_asset_ids = batch_importer.import_batch(batch_assets, verify)
+            created_asset_ids.append(batch_created_asset_ids)
```
PierreLeveau, line 476: extend instead  of append? We want a flat list imho

PierreLeveau: maybe worth doing a test by mocking the resolver answer

PierreLeveau (changes requested):

src/kili/services/asset_import/base.py
theodu: yes you are right!

src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -140,14 +140,14 @@ def append_many_to_dataset(
...
-        result = import_assets(
+        created_assets_ids = import_assets(
             self,
             project_id=project_id,
             assets=assets,
             disable_tqdm=disable_tqdm,
             verify=wait_until_availability,
         )
-        return result
+        return {"id": project_id, "created_assets_ids": created_assets_ids}
```
Jonas1312, line 150: I would just name the key `asset_ids`

src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -50,7 +50,7 @@ def append_many_to_dataset(
         json_metadata_array: Optional[List[dict]] = None,
         disable_tqdm: bool = False,
         wait_until_availability: bool = True,
-    ) -> Optional[Dict[Literal["id"], str]]:
+    ) -> Optional[Dict[Literal["id", "created_assets_ids"], Union[str, List[str]]]]:
         # pylint: disable=line-too-long
         """Append assets to a project.
```
Jonas1312, line 55: need to change the docstring `Returns: ...`

src/kili/services/asset_import/base.py
```diff
@@ -86,16 +87,20 @@ def __init__(self, kili, project_params: ProjectParams, batch_params: BatchParam
         self.logger.setLevel(logging.INFO)
 
     @pagination.api_throttle
```
Jonas1312, line 89: I thought I had removed all api_throttle, maybe you need to rebase on main

src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -140,14 +140,14 @@ def append_many_to_dataset(
         for key, value in field_mapping.items():
             if value is not None:
                 assets = [{**assets[i], key: value[i]} for i in range(nb_data)]
-        result = import_assets(
+        created_assets_ids = import_assets(
```
Jonas1312, line 143: so for async imports the asset_ids would be empty?  if yes, we should put it in the docstring

Jonas1312 (changes requested):

src/kili/entrypoints/mutations/asset/__init__.py
theodu: it was to make sure that the client understands that we don't return all the project asset ids

theodu: But I am also good with just asset_ids

Jonas1312: ah ok makes sense, you can keep it as is then

theodu: I used asset_ids, I also hesitated to put it as so

Jonas1312 (approved):

tests/services/asset_import/test_import_common.py
```diff
@@ -55,6 +57,21 @@ def test_cannot_upload_raw_text_to_pdf_project(self, *_):
         with self.assertRaises(FileNotFoundError):
             import_assets(self.kili, self.project_id, assets, disable_tqdm=True)
 
+    @patch.object(ProjectQuery, "__call__", side_effect=mocked_project_input_type("TEXT"))
+    @patch.object(AssetQuery, "count", return_value=1)
+    def test_return_the_ids_of_created_assets(self, *_):
+        assets = [{"content": "One"}, {"content": "Two"}, {"content": "Three"}]
+
+        def graphql_execute_side_effect(*args, **kwargs):
```
PierreLeveau, line 65: this is a hack until we get the new architecture in place correct?

theodu: yes, it will be much easier after!

PierreLeveau (approved):

PR #1360, opened 2023-07-28
fix: rate limiting at graphql client level

Jonas1312:


theodu (approved):
Wonderful! super improvement!

PR #1376, opened 2023-08-03
feat: add overwritting of prediction labels when appending labels

theodu:


src/kili/entrypoints/mutations/label/__init__.py
```diff
@@ -44,6 +44,7 @@ def create_predictions(
         model_name: Optional[str] = None,
         asset_id_array: Optional[List[str]] = None,
         disable_tqdm: bool = False,
+        overwrite: bool = True,
```
Jonas1312, line 47: I find it risky to change the current default behavior

On the "Explore" UI, it's nice to be able to see the different predictions made by a model over time for example

I would set it to false

src/kili/entrypoints/mutations/label/__init__.py
```diff
@@ -57,6 +58,9 @@ def create_predictions(
             model_name: The name of the model that generated the predictions
             asset_id_array: The internal IDs of the assets for which we want to add predictions.
             disable_tqdm: Disable tqdm progress bar.
+            overwrite: when uploading predictions or inference labels, if True,
+                it will overwrite existing predictions and inference labels
```
Jonas1312, line 62: remove "s" at predictions

Jonas1312, line 62: so here we are in the create_predictions method, so we shouldn't even be talking about inference labels here

src/kili/entrypoints/mutations/label/__init__.py
```diff
@@ -206,6 +211,9 @@ def append_labels(  # pylint: disable=dangerous-default-value
             project_id: Identifier of the project.
             asset_external_id_array: list of asset external ids to append labels on.
             disable_tqdm: Disable tqdm progress bar.
+            overwrite: when uploading predictions or inference labels, if True,
+                it will overwrite existing predictions and inference labels
+                with the same model name on the targeted assets.
```
Jonas1312, line 216: same comments as above

also, I don't understand the "it will overwrite existing predictions and inference labels".

Should we really overwrite both prediction and inference labels if we call kili.append_labels(label_type="prediction")?

I think we should only delete existing labels of the same type

If a user wants to delete both, he can call kili.append_labels a second time with the other label type

wdyt?

Jonas1312 (changes requested):

src/kili/entrypoints/mutations/label/__init__.py
theodu: I agree with you indeed

src/kili/entrypoints/mutations/label/__init__.py
theodu: yes it is the current behavior, it is the docstrings that doesn't explain it well indeed

src/kili/entrypoints/mutations/label/__init__.py
```diff
@@ -206,6 +211,8 @@ def append_labels(  # pylint: disable=dangerous-default-value
             project_id: Identifier of the project.
             asset_external_id_array: list of asset external ids to append labels on.
             disable_tqdm: Disable tqdm progress bar.
+            overwrite: if True, it will overwrite existing predictions of
```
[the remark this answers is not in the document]
theodu: ah yeah indeed, I think that I switched all the 3 functions, good catch

src/kili/entrypoints/mutations/label/__init__.py
```diff
@@ -206,6 +210,9 @@ def append_labels(  # pylint: disable=dangerous-default-value
             project_id: Identifier of the project.
             asset_external_id_array: list of asset external ids to append labels on.
             disable_tqdm: Disable tqdm progress bar.
+            overwrite: when uploading prediction or inference labels, if True,
+                it will overwrite existing labels with the same model name
+                and of the same type label type, on the targeted assets.
```
Jonas1312, line 215: same type label type

Jonas1312 (approved):

PR #1377, opened 2023-08-03
fix: disable urllib3 warnings related to connection pool

Jonas1312:
I was spammed with those warnings when adding assets to a kili project:

```
ARNING:urllib3.connectionpool:Connection pool is full, discarding connection: storage.googleapis.com. Connection pool size: 10
WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: storage.googleapis.com. Connection pool size: 10
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 47/47 [00:06<00:00,  7.64it/s]
  0%|                                                                                                                                                       | 0/11 [00:01<?, ?it/s]WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: storage.googleapis.com. Connection pool size: 10
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 11/11 [00:03<00:00,  3.09it/s]
  0%|                                                                                                                                                       | 0/14 [00:01<?, ?it/s]WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: storage.googleapis.com. Connection pool size: 10
WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: storage.googleapis.com. Connection pool size: 10
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 14/14 [00:03<00:00,  4.09it/s]
  0%|                                                                                                                                                       | 0/20 [00:01<?, ?it/s]WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: storage.googleapis.com. Connection pool size: 10
WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: storage.googleapis.com. Connection pool size: 10
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 20/20 [00:02<00:00,  6.86it/s]
  0%|                                                                                                                                                       | 0/15 [00:01<?, ?it/s]WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: storage.googleapis.com. Connection pool size: 10
```

### DIFF
--- a/src/kili/client.py
+++ b/src/kili/client.py
@@ -1,4 +1,5 @@
 """This script permits to initialize the Kili Python SDK client."""
+import logging
 import os
 import warnings
 from datetime import datetime, timedelta
@@ -37,6 +38,17 @@ from kili.exceptions import AuthenticationFailed, UserNotFoundError
 from kili.internal import KiliInternal
 
 warnings.filterwarnings("default", module="kili", category=DeprecationWarning)
+
+
+class FilterPoolFullWarning(logging.Filter):
+    """Filter out the specific urllib3 warning related to the connection pool."""
+
+    def filter(self, record) -> bool:
+        """urllib3.connectionpool:Connection pool is full, discarding connection: ..."""
+        return "Connection pool is full, discarding connection" not in record.getMessage()
+
+
+logging.getLogger("urllib3.connectionpool").addFilter(FilterPoolFullWarning())
 
 
 class Kili(  # pylint: disable=too-many-ancestors,too-many-instance-attributes


theodu:
yeah indeed I had them too since yesterday.
Did you understand what was happening exactly ?

Jonas1312:
> yeah indeed I had them too since yesterday. Did you understand what was happening exactly ?

Apparently good news is  that no data is lost

I think it's due  to the fact that we now use a common requests.Session for all http requests done in our repo (before we would open a session for each request)

So in a requests.Session is assigned an requests HTTP adapter, which has a connection pool of size 10 by default

urllib3 throws this error because the connection that just finished will be discarded after the request is completed (because the pool is full, as mentioned). This means that this particular connection is not going to be re-used in the future.

Because a urllib3 PoolManager reuses connections, it will limit how many connections are retained per host to avoid accumulating too many unused sockets. 

https://stackoverflow.com/questions/53765366/urllib3-connectionpool-connection-pool-is-full-discarding-connection

src/kili/client.py
```diff
@@ -39,6 +40,17 @@
...
+    """Filter out the specific urllib3 warning related to the connection pool."""
+
+    def filter(self, record) -> bool:
+        """urllib3.connectionpool:Connection pool is full, discarding connection: ..."""
+        return "Connection pool is full, discarding connection" not in record.getMessage()
+
+
+logging.getLogger("urllib3.connectionpool").addFilter(FilterPoolFullWarning())
+
+
```
theodu, line 53: and creating a class like this which is not called anywhere is enough for filtering ?

Jonas1312: it worked for me, can  you try it too? Since you had the same issue apparently

theodu (approved):

src/kili/client.py
theodu: works for me too

PR #1380, opened 2023-08-04
chore(DATA-1438): 🤖 remove deprecated "issueNumber" from default fields

gheorghetutunaru:


Jonas1312 (commented):
should we remove the others too?

image

gheorghetutunaru:
> should we remove the others too?
> 
No, it is sill considered as `obligatory` when creating the issues. But maybe we should make it optional, @p-desaintchamas do you agree ?

Jonas1312 (approved):

MicrexIT (approved):

PR #1382, opened 2023-08-09
feat: ask for api key if not provided when init Kili client

Jonas1312:


src/kili/client.py
```diff
@@ -124,16 +125,18 @@ def __init__(
             kili.projects()  # list your projects
             ```
         """
-        if api_key is None:
-            api_key = os.getenv("KILI_API_KEY")
+        api_key = api_key or os.getenv("KILI_API_KEY")
+
+        if not api_key:
+            api_key = getpass.getpass("Please enter your API key: ")
```
theodu, line 131: very cool idea!
Could we also display a message explaining that the user can export it as an env var to avoid doing it each time ?

theodu (approved):

PR #1388, opened 2023-08-17
feat:added pascalvoc files

manuelcgallucci:


theodu:
can't these images be found with an url ? like in the COCO tutorial ?
Or is it only possible to download the whole dataset ?

manuelcgallucci:
It is only possible to download the whole dataset, there are no links for individual images as in the COCO dataset. Thats why i only added a few images

theodu (approved):

PR #1389, opened 2023-08-17
feat: freeze linting dependencies

theodu:


setup.py
```diff
@@ -54,11 +54,11 @@
...
-        "black",
-        "pre-commit",
-        "pylint",
-        "flake8-unused-arguments",
-        "pyright",
+        "black>=23.7.0,<24.0.0",
+        "pre-commit>=3.3.0,<4.0.0",
+        "pylint==2.17.5",
+        "flake8-unused-arguments==0.0.13",
+        "pyright==1.1.323",
```
Jonas1312, line 61: why not lock major versions of those?

Jonas1312 (approved):

setup.py
PierreLeveau: because even in patch versions there maybe new error detected (happened).

setup.py
```diff
@@ -54,11 +54,11 @@
         "mike",
         "pymdown-extensions",
         # linting
-        "black",
-        "pre-commit",
-        "pylint",
-        "flake8-unused-arguments",
-        "pyright",
+        "black>=23.7.0,<24.0.0",
```
PierreLeveau, line 57: for black, the source of truth is the one in the precommit config. To keep the formatting consistent between vscode and the precommits, we may ensure that the versions are in sync (out of the scope of this PR).

setup.py
PierreLeveau: for example:`pyright==1.1.316` on the branch release/2.129.0 finds errors, `pyright==1.1.300` does not.

PierreLeveau (approved):

PR #1391, opened 2023-08-21
test: reorganize tests folder

Jonas1312:
I encountered a weird bug... Some kili client tests were not passing anymore: 

```text
>       _ = Kili(api_key="fake_api_key")
E       TypeError: object.__new__() takes exactly one argument (the type to instantiate)
```

Turns out it was some CLI integrations tests mocking the `Kili.__new__` that was creating this bug.

I had to modify many CLI tests to avoid mocking `__new__`


.github/workflows/ci.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           pip install -e ".[dev]"
 
       - name: Unit tests
```
theodu, line 82: can you rename it "Unit & Integration tests" ?

theodu: jut above too in the job name too

theodu (approved):

PR #1397, opened 2023-08-22
fix: add `status_in` filter to projectUsers query

Jonas1312:


theodu (commented):
It is great to now be able to filter on status. 
however there is not quick way to get ACTIVATED and ORG_ADMIN users. When I created the ticket, my idea was to have a way to quickly get only people active in the project (activated + org_admin) to be able to iterate on them. Is it what the activeInProject is for ? Could it lead to another method argument ?

PierreLeveau (approved):

Jonas1312:
> It is great to now be able to filter on status. however there is not quick way to get ACTIVATED and ORG_ADMIN users. When I created the ticket, my idea was to have a way to quickly get only people active in the project (activated + org_admin) to be able to iterate on them. Is it what the activeInProject is for ? Could it lead to another method argument ?

no activeInProject is used for detecting active users in project, that have labeled some assets etc

The implementation follows the graphql schema where we can only have one value for status:
```graphql
input ProjectUserWhere {
  id: ID
  status: ProjectUserStatus
  isAuthor: Boolean
  project: ProjectWhere
  user: UserWhere
  activeInProject: Boolean
}
```

so can query twice in our SDK implementation to have both org_admin and active project users, or we let the user call this method twice

theodu (approved):

theodu:
> I set the default status_in to ("ACTIVATED", "ORG_ADMIN"), even though IMO we should use the same behavior as in the UI, that is ("ACTIVATED", )

Indeed, i don't know which one is the best.

theodu:
we should take an implem on the backend too to implement status_in later too. I think that it would have required as much work as in the SDK and would have been cleaner

PR #1415, opened 2023-08-28
feat: add a first query to the new repo archi

theodu:


src/kili/core/helpers.py
```diff
@@ -328,18 +328,6 @@ def check_file_mime_type(path: str, input_type: str, raise_error=True) -> bool:
     return correct_mime_type
 
 
-def disable_tqdm_if_as_generator(as_generator: bool, disable_tqdm: bool):
```
theodu, line 331: moved to presentation/common_validator.py module

src/kili/core/enums.py
```diff
@@ -29,16 +29,6 @@
 ]
 
 
-InputType = Literal[
```
theodu, line 32: moved to domain/project.py module

src/kili/core/graphql/__init__.py
```diff
@@ -1,28 +1,21 @@
...
 from kili.core.helpers import format_result
+from kili.gateways.kili_api_gateway.queries import QueryOptions
 from kili.utils.tqdm import tqdm
 
 from .graphql_client import GraphQLClient
 
 T = TypeVar("T")
 
 
-class QueryOptions(NamedTuple):
```
theodu, line 18: moved to gateway/kili_api_gateway/queries.py module

src/kili/use_cases/asset/exceptions.py
theodu, line 1: Question here: where should exceptions be ? Should they be split in the code, close to where there are used or should they all be in an `exceptions.py` module at the root of the project to be imported by clients to do try except in his script ?

Jonas1312:
@theodu can you fix the conflicts before we review?  thx

src/kili/gateways/kili_api_gateway/asset/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+            self.graphql_client, where, options, GQL_COUNT_ASSETS
+        )
+        return PaginatedGraphQLQuery(self.graphql_client).execute_query_from_paginated_call(
+            query, where, options, nb_elements_to_query, post_call_function
+        )
+
+    def count_assets(self, where: AssetWhere):
+        """Send a GraphQL request calling countIssues resolver."""
+        payload = {
+            "where": where.build_gql_value(),
```
Jonas1312, line 46: why value? should be build_gql_where no?

src/kili/gateways/kili_api_gateway/asset/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+        fragment = fragment_builder(fields)
+        query = get_asset_query(fragment)
+        nb_elements_to_query = get_number_of_elements_to_query(
+            self.graphql_client, where, options, GQL_COUNT_ASSETS
+        )
+        return PaginatedGraphQLQuery(self.graphql_client).execute_query_from_paginated_call(
+            query, where, options, nb_elements_to_query, post_call_function
+        )
+
+    def count_assets(self, where: AssetWhere):
```
Jonas1312, line 43: type hint

src/kili/gateways/kili_api_gateway/project/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+class ProjectOperationMixin:
+    """Mixin extending Kili API Gateway class with Assets related operations."""
+
+    graphql_client: GraphQLClient
+
+    def get_project(
+        self,
+        project_id,
+        fields: List[str],
+    ):
```
Jonas1312, line 22: type hints

src/kili/gateways/kili_api_gateway/asset/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+        fields: List[str],
+        options: QueryOptions,
+        post_call_function: Optional[Callable],
+    ):
+        """List assets with given options."""
+        fragment = fragment_builder(fields)
+        query = get_asset_query(fragment)
+        nb_elements_to_query = get_number_of_elements_to_query(
+            self.graphql_client, where, options, GQL_COUNT_ASSETS
+        )
```
Jonas1312, line 38: is it needed? There is a `_count` in the PaginatedGraphQLQuery already

src/kili/gateways/kili_api_gateway/project/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+        """List assets with given options."""
+        fragment = fragment_builder(fields)
+        query = get_project_query(fragment)
+        where = ProjectWhere(project_id=project_id)
+        result = self.graphql_client.execute(
+            query=query, variables={"where": where.build_gql_value()}
```
Jonas1312, line 28: first=1

tests/unit/services/export/test_geojson.py
```diff
@@ -18,10 +18,10 @@ def test_kili_export_labels_geojson(mocker: pytest_mock.MockerFixture):
...
     }
     mocker.patch("kili.services.export.get_project", return_value=get_project_return_val)
-    mocker.patch(
-        "kili.entrypoints.queries.asset.media_downloader.ProjectQuery.__call__",
-        return_value=(i for i in [get_project_return_val]),
-    )
+    # mocker.patch(
+    #     "kili.entrypoints.queries.asset.media_downloader.ProjectQuery.__call__",
+    #     return_value=(i for i in [get_project_return_val]),
+    # )
```
Jonas1312, line 24: comments

tests/unit/services/label_data_parsing/test_label_utils.py
```diff
@@ -261,19 +260,6 @@ def test_integration_of_label_parsing_in_kili_labels_assert_types(mocker):
     )
 
 
-def test_integration_of_label_parsing_in_kili_asset_assert_types(mocker):
```
Jonas1312, line 264: where have those tests been moved to?

Jonas1312 (changes requested):

tests/unit/services/label_data_parsing/test_label_utils.py
theodu: So the test about the return types have been integrated to in `tests/integration/entrypoints/client/queries/test_user_facing_queries.py::test_assets_query_return_type` because there were already a test on the asset query return type based on its input arguments.

and the test about the parsing when calling the function has been moved to `tests/integration/use_cases/test_asset.py` because it is not a unit but an integration test. We don't test the parsing class but the fact that when calling the use case, it will go though the parsing process

src/kili/gateways/kili_api_gateway/asset/__init__.py
theodu: ah well spotted,` _count` is a dead code and not called anywhere else. i delete it.

In order to make things easier and not to dependant, I moved the count query outside of the `PaginatedGraphQLQuery`. if one query has a count method and that it could provide a better user experience with `tqdm` logging, we can execute this query in the use case and pass it to the `execute_query_from_paginated_call` function for it to be aware of the number of elements it has to retrieve. If a query does not have an associated count query, then we can still paginate but without progress logging

tests/unit/services/label_data_parsing/test_label_utils.py
Jonas1312: tests/integration/entrypoints/client/queries/test_user_facing_queries.py is not here

theodu: -> `tests/integration/entrypoints/client/queries/test_assets_queries.py`
i don't know where I found test_user_facing_queries

src/kili/use_cases/asset/exceptions.py
Jonas1312: I think we should declare and implement the exceptions close to where they are used

But we can also expose them all in a kili.exceptions module so that they can be imported easily by users

src/kili/gateways/kili_api_gateway/queries.py
```diff
@@ -0,0 +1,164 @@
...
+            # add the subquery to the fragment
+            fragment += f" {root_field}{{{new_fragment}}}"
+
+        # remove the fields that have been queried in subqueries (e.g. "roles.user.id")
+        fields = [field for field in fields if "." not in field]
+
+    for field in fields:
+        fragment += f" {field}"
+
+    return fragment
```
Jonas1312, line 164: you moved fragment_builder here but didn't remove it from `class GraphQLQuery(ABC)`, we should avoid duplicating code

this method is can be removed from the class, it doesn't really need the `self` anyway

theodu: good idea

src/kili/gateways/kili_api_gateway/asset/__init__.py
theodu: changed to build_gql_where

src/kili/domain/project.py
```diff
@@ -0,0 +1,20 @@
...
+    "VIDEO",
+    "VIDEO_LEGACY",
+]
+
+
+@dataclass
+class Project:
+    """Project Entity."""
+
+    id_: str
```
Jonas1312, line 20: we're not using it anywhere

src/kili/domain/asset.py
```diff
@@ -0,0 +1,9 @@
+"""Asset domain."""
+from dataclasses import dataclass
+
+
+@dataclass
+class Asset:
+    """Asset Entity."""
+
+    id_: str
```
Jonas1312, line 9: we're not using it anywhere

Jonas1312 (commented):
for entities like this:

```python
@dataclass
class SomeEntity:
    """Some Entity."""
    id_: str
```

They are mostly aimed at typing the return of the graphql query, which is quite often just an id. So just a `IssueId = NewType("IssueId ", str)` is enough IMO

theodu:
> for entities like this:
> 
> ```python
> @dataclass
> class SomeEntity:
>     """Some Entity."""
>     id_: str
> ```
> 
> They are mostly aimed at typing the return of the graphql query, which is quite often just an id. So just a `IssueId = NewType("IssueId ", str)` is enough IMO

I agree with you, it was one of the point that I wanted to see with you. No need to define entities class I think because we don't use them in queries and we only use ids in mutations

src/kili/adapters/kili_api_gateway/project/types.py
```diff
@@ -0,0 +1,29 @@
...
+        """Build the GraphQL Where payload sent in the resolver from the SDK ProjectWhere."""
+        value = {
+            "id": self.project_id,
+            "searchQuery": self.search_query,
+            "updatedAtGte": self.updated_at_gte,
+            "updatedAtLte": self.updated_at_lte,
+        }
+        if self.archived is not None:
+            value["archived"] = self.archived  # type: ignore
+        return value
```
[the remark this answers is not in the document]
theodu: I remove this one because I don't need it for the moment

src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -0,0 +1,165 @@
...
+    """Abtsract class for defining the where payload to send in a graphQL query."""
+
+    @abstractmethod
+    def build_gql_where(self):
+        """Build the GraphQL where variable sent in the resolver from the
+        arguments given to the where class."""
+        raise NotImplementedError
+
+
+class PaginatedGraphQLQuery:
```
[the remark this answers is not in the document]
theodu: yes there are two of them but it is the time to migrate other queries..
The goal is to soon delete the old graphQL Query

src/kili/adapters/kili_api_gateway/project/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+        result = self.graphql_client.execute(
+            query=query, variables={"where": where.build_gql_where(), "first": 1}
+        )
+        projects = result["data"]
+        if len(projects) == 0:
+            raise NotFound(
+                f"project ID: {project_id}. Maybe your KILI_API_KEY does not belong to a member of"
+                " the project."
+            )
+        return projects[0]
```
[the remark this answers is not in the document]
theodu: I removed the service since it is not necessary anymore

Jonas1312 (approved):
good once e2e tests pass

PierreLeveau (approved):

PierreLeveau (approved):

PR #1416, opened 2023-08-29
feat: batch question creation

theodu:


src/kili/gateways/kili_api_gateway/issue/__init__.py
```diff
@@ -26,7 +26,7 @@ def create_issues(
     ) -> List[Issue]:
         """Send a GraphQL request calling createIssues resolver."""
         created_issue_entities: List[Issue] = []
-        with tqdm.tqdm(total=len(issues)) as pbar:
+        with tqdm.tqdm(total=len(issues), desc="Creating issues") as pbar:
```
theodu, line 29: I also added the message on issue creation, it gives a nicer tqdm bar with the description. Very user-friendly, we should use it from now on.
Capture d’écran 2023-08-29 à 10 52 54

PierreLeveau: I agree.

PierreLeveau (approved):

PR #1419, opened 2023-08-30
feat: add `kili.tags` and `kili.tag_project` methods

Jonas1312:
for kili.tag_project, I accept both tag ids and tag labels in the `tags` parameter. This assumes that nobody would create a tag label named after a tag id...

Not sure if we should accept tag ids at all? If yes, maybe add a second argument `tag_ids`?

wdyt?


src/kili/presentation/client/tag.py
```diff
@@ -0,0 +1,75 @@
...
+from kili.utils.logcontext import for_all_methods, log_call
+from kili.utils.tqdm import tqdm
+
+
+@for_all_methods(log_call, exclude=["__init__"])
+class TagClientMethods:
+    """Methods attached to the Kili client, to run actions on tags."""
+
+    kili_api_gateway: KiliAPIGateway
+    http_client: requests.Session
```
PierreLeveau, line 18: not used here

PierreLeveau (changes requested):

src/kili/gateways/kili_api_gateway/tag/operations.py
```diff
@@ -0,0 +1,9 @@
+"""GraphQL Tags operations."""
+
+GQL_CHECK_TAG = """
```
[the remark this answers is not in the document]
Jonas1312: I put all queries in the operations.py. So that we can reuse them in the tests

theodu:
waiting for the rebase to review

Jonas1312:
I'm still not sure if we should accept tag ids or tag labels as parameters of the tag methods

src/kili/use_cases/issue/__init__.py
```diff
@@ -4,17 +4,20 @@
...
     """Issue use cases."""
 
-    def __init__(self, kili_api_gateway: KiliAPIGateway):
+    def __init__(self, kili_api_gateway: KiliAPIGateway) -> None:
         self._kili_api_gateway = kili_api_gateway
 
-    def create_issues(self, project_id, issues: List[IssueToCreateUseCaseInput]):
+    def create_issues(
+        self, project_id: str, issues: List[IssueToCreateUseCaseInput]
+    ) -> List[IssueId]:
```
theodu, line 20: It is better to force the return type hint even if it is well infered from the function implem with inputs ?

theodu:
> I'm still not sure if we should accept tag ids or tag labels as parameters of the tag methods

good question, I think that is very good to be able to provide both (as you did). however, I wonder what is the best between having one field to provide both or to separate them into two fields for the user.

Jonas1312:
> > I'm still not sure if we should accept tag ids or tag labels as parameters of the tag methods
> 
> good question, I think that is very good to be able to provide both (as you did). however, I wonder what is the best between having one field to provide both or to separate them into two fields for the user.

yes maybe two separate parameters would be safer, wdyt @PierreLeveau

tests/integration/use_cases/test_tag.py
```diff
@@ -0,0 +1,55 @@
...
+    # When
+    fetched_tags = TagUseCases(kili_api_gateway).get_tags_of_project(
+        project_id="fake_proj_id", fields=["id", "label"]
+    )
+
+    # Then
+    assert fetched_tags == tags
+
+
+def test_when_tagging_project_then_it_tags_the_project(kili_api_gateway: KiliAPIGateway):
```
theodu, line 37: we can add a test to assert that it raises an error when a tag is not found in the orga tags

src/kili/use_cases/issue/__init__.py
Jonas1312: Doesn't cost much to add it. I find that it helps, especially during refactoring

src/kili/adapters/kili_api_gateway/project/__init__.py
```diff
@@ -5,15 +5,14 @@
...
 from kili.domain.project import ProjectId
 from kili.exceptions import NotFound
 
+from ..base import BaseOperationMixin
 
-class ProjectOperationMixin:
-    """Mixin extending Kili API Gateway class with Assets related operations."""
 
-    graphql_client: GraphQLClient
+class ProjectOperationMixin(BaseOperationMixin):
```
PierreLeveau, line 14: with Project related operations

PierreLeveau (approved):

PierreLeveau (changes requested):
add separate field for tag id and tag name

src/kili/adapters/kili_api_gateway/project/__init__.py
```diff
@@ -5,15 +5,14 @@
...
+    """Mixin extending Kili API Gateway class with Assets related operations."""
```
PierreLeveau, line 15: with Projects related operations

PR #1421, opened 2023-09-01
feat: 🎸 update mkdoc for plugins

florianlega:


p-desaintchamas (approved):

src/kili/services/plugins/model.py
```diff
@@ -148,15 +148,21 @@ def on_project_updated(
...
 
         !!! example
             ```python
-            def on_project_updated(self, label: List[Dict]):
-                for annotation in annotations
-                    key = annotation.get('key')
-                    if key == 'name':
-                        save_name(annotation.get('newValue'))
+            def on_project_updated(self, settings_updated: List[Dict]):
+                for setting in settings_updated
```
Jonas1312, line 162:
```suggestion
                for setting in settings_updated:
```

Jonas1312 (approved):

src/kili/services/plugins/model.py
```diff
@@ -148,15 +148,21 @@ def on_project_updated(
         Args:
             settings_updated: Settings updated on the project a list of
                 dictionary containing the following fields:
-                `key`, `newValue`, `oldValue`.
+                `key`, `newValue`, `oldValue`few.
```
p-desaintchamas, line 151: extra few ?

p-desaintchamas:
```suggestion
                `key`, `newValue`, `oldValue`.
```

PR #1425, opened 2023-09-04
fix: file extension guess when encoding parameter is in the content type

theodu:


Jonas1312 (changes requested):
we should add a test or a comment to understand why

Jonas1312 (approved):

PR #1430, opened 2023-09-05
feat: add `kili.untag_project`

Jonas1312:


src/kili/core/graphql/graphql_client.py
```diff
@@ -90,7 +90,7 @@ def __init__(
             use_json=True,
             timeout=30,
             verify=verify,
-            retries=20,
+            # retries=20,  # FIXME: when it is set, it sometimes freezes the .execute() method
```
PierreLeveau, line 93: by default, this won't retry at all?

PierreLeveau (changes requested):

src/kili/core/graphql/graphql_client.py
Jonas1312: I investigated the issue yesterday, and it appears that when we send an invalid query to the backend (for example we forget to provide a mandatory parameter in a graphql query), the backend will return the error with a 500 HTTP status code.

So here our graphql client (that relies on requests) will retry again and again even though the server answered.

I excluded the 500 code from the retry_status_forcelist, but other 500 errors raised by the backend might make the SDK crash

PierreLeveau (approved):

PR #1431, opened 2023-09-05
feat: add `tags` parameter to `kili.create_project`

Jonas1312:


src/kili/entrypoints/mutations/project/helpers.py
```diff
@@ -8,9 +8,11 @@ def verify_argument_ranges(
...
     min_consensus_size: Optional[int],
     review_coverage: Optional[int],
-):
-    """Ensure that all arguments are set in a correct range or raise
-    a ValueError otherwise."""
+) -> None:
+    """Ensure that all arguments are set in a correct range.
+
+    Raise a ValueError otherwise.
+    """
```
theodu, line 15: just to inform because it is not a big deal: the google style doctrings has a field dedicated to error raised: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
```
"""
Func Description

Args:

Returns:

Raises:
"""
```

src/kili/presentation/client/project.py
```diff
@@ -0,0 +1,85 @@
...
+            input_type=input_type,
+            json_interface=json_interface,
+            title=title,
+            description=description,
+            project_type=project_type,
+        )
+
+        if tags is not None:
+            tag_use_cases = TagUseCases(self.kili_api_gateway)
+            tag_ids = tag_use_cases.get_tag_ids_from_labels(labels=tags)
```
theodu, line 82: this one should rather be a function in the gateway instead of a usecase

theodu: but this will become useless because you plan to allow to give tags label in tag_project right ?

Jonas1312: > this one should rather be a function in the gateway instead of a usecase

why?



> but this will become useless because you plan to allow to give tags label in tag_project right ?

it's easier to implement and test two use cases than a big one I think

theodu: yes but it is not a use case, it is just a way to get infos needed for the use case from Kili.

theodu (approved):

tests/integration/use_cases/test_project.py
```diff
@@ -0,0 +1,18 @@
...
+    project_id = ProjectUseCases(kili_api_gateway).create_project(
+        input_type="TEXT",
+        json_interface={},
+        title="test",
+        description="description",
+        project_type=None,
+    )
+
+    # Then
+    assert project_id == "fake_project_id"
```
PierreLeveau, line 18: how does it play with that? since it is mocked , the gateway it should not be able to fetch the project then return
```for attempt in Retrying(
            stop=stop_after_delay(60),
            wait=wait_fixed(1),
            retry=retry_if_exception_type(NotFound),
            reraise=True,
        ):
            with attempt:
                _ = self._kili_api_gateway.get_project(project_id=project_id, fields=["id"])
                ```

Comment (Jonas1312) [reply tests/integration/use_cases/test_project.py:18]:
```python
_ = self._kili_api_gateway.get_project(project_id=project_id, fields=["id"])
```

will return a magic mock, so it won't raise anything, so tenacity won't retry

PR #1439, opened 2023-09-05
fix: formating of paginated queries

theodu:


src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -34,8 +36,9 @@ class PaginatedGraphQLQuery:
     It factorizes code for executing paginated queries.
     """
 
-    def __init__(self, graphql_client: GraphQLClient):
+    def __init__(self, graphql_client: GraphQLClient, http_client: HttpClient):
         self._graphql_client = graphql_client
+        self.http_client = http_client
```
Jonas1312, line 41: private member

src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -82,8 +85,13 @@ def execute_query_from_paginated_call(
                     )
                     payload = {"where": where, "skip": skip, "first": first}
                     elements = self._graphql_client.execute(query, payload)["data"]
+                    assert isinstance(
+                        elements, list
+                    ), "Paginated queries does not support operations that does not return a list"
+                    print(elements)
```
Jonas1312, line 91: print

src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -82,8 +85,13 @@ def execute_query_from_paginated_call(
...
+                    elements: List = [format_json(elem, self.http_client) for elem in elements]
 
-                    if elements is None or len(elements) == 0:
+                    if len(elements) == 0:
```
Jonas1312, line 94: is it guaranteed that elements can't be None?

src/kili/core/helpers.py
```diff
@@ -86,15 +86,20 @@ def is_url(path: object):
     return isinstance(path, str) and re.match(r"^(http://|https://)", path.lower())
 
 
-def format_json_dict(result: Dict, http_client: HttpClient) -> Dict:
-    """Formats the dict part of a json return by a GraphQL query into a python object.
+def __format_json_dict(result: Dict, http_client: HttpClient) -> Dict:
```
Jonas1312, line 89: why double underscore?

Jonas1312: I also thought we wanted to get rid of this ugly format_result/format_json/format_json_dict stuff?

tests/unit/test_paginated_graphql_query.py
```diff
@@ -33,25 +35,34 @@ def mocked_client_execute(_, payload):
     return mocked_graphql_client
 
 
-def test_given_a_query_the_function_returns_a_generator(graphql_client: GraphQLClient):
+@pytest.fixture
+def http_client() -> HttpClient:
+    return MagicMock(spec=HttpClient)
```
Jonas1312, line 40: I would put this in conftest, since it's a general fixture

tests/unit/test_paginated_graphql_query.py
```diff
@@ -151,18 +164,112 @@ def test_given_a_query_and_a_number_of_elements_to_query_i_have_a_progress_bar(
...
+
+
+def test_given_a_query_returning_serialized_json_it_parses_json_fields(
+    http_client,
+):
+    # given
+    options = QueryOptions(disable_tqdm=False)
+    number_of_elements_to_query = 1
+    post_call_function = None
+    graphql_client = MagicMock(spec=GraphQLClient)
```
Jonas1312, line 191: we could make this a fixture too, and put it in conftest

tests/conftest.py
```diff
@@ -8,6 +8,15 @@
 from kili.core.graphql.graphql_client import GraphQLClient
 
 
+class MockResponse:
```
Jonas1312, line 11: MockRequestsReponse

and add type hints

PierreLeveau: this is not a Mock but a Fake

PierreLeveau (approved):

src/kili/domain/asset.py
```diff
@@ -44,3 +44,6 @@ class AssetFilters:
     inference_mark_lte: Optional[float] = None
     issue_type: Optional[IssueType] = None
     issue_status: Optional[IssueStatus] = None
+
+
+ASSET_JSON_FIELDS = ["jsonMetadata", "labels.jsonResponse", "latestLabel.jsonResponse"]
```
PierreLeveau, line 49: to put in the GateWay, we do not need them in the domain

setup.py
```diff
@@ -25,6 +25,7 @@
     "filelock>=3.0.0,<4.0.0",
     "pip-system-certs>=4.0.0,<5.0.0; platform_system=='Windows'",
     "pyrate-limiter>=2,<3",
+    "more-itertools>=8.0.0,<11.0.0",
```
PierreLeveau, line 28: we only need one method that can be replaced by:
```
from itertools import islice

def batcher(iterable, batch_size):
    iterator = iter(iterable)
    while batch := list(islice(iterator, batch_size)):
        yield batch
```
not sure it is necessary to add another lib

tests/unit/test_paginated_graphql_query.py
```diff
@@ -151,18 +164,112 @@ def test_given_a_query_and_a_number_of_elements_to_query_i_have_a_progress_bar(
...
+    }
+
+    # when
+    gen = PaginatedGraphQLQuery(graphql_client, http_client).execute_query_from_paginated_call(
+        QUERY, WHERE, options, "", number_of_elements_to_query, post_call_function
+    )
+    assets = list(gen)
+
+    # then
+    assert assets == [
```
PierreLeveau, line 210: or 
```
assert next(gen) = {"jsonMetadata": ....}
```

tests/unit/test_paginated_graphql_query.py
```diff
@@ -151,18 +164,112 @@ def test_given_a_query_and_a_number_of_elements_to_query_i_have_a_progress_bar(
...
```
PierreLeveau, line 247: same as above

PierreLeveau (changes requested):

src/kili/adapters/kili_api_gateway/helpers/queries.py
theodu: yes by the isinstance(elements, list) just above

src/kili/core/helpers.py
theodu: double underscore prevent the function to be imported in another file. It is to avoid this function to be reused somewhere else.

theodu: I don't make a PR to remove this format result. it is still used in many queries and mutations. We will be able to simply delete it once we will have moved every function. We just have to make sure not to use it anymore

tests/unit/test_paginated_graphql_query.py
theodu: this is already a fixture in the conftest. 
But there is already another graphql_client fixture in the file to override the base on in the conftest so I need to redefine it here

setup.py
theodu: Ah great!

src/kili/adapters/kili_api_gateway/asset/formatters.py
```diff
@@ -0,0 +1,19 @@
+"""Formatters for assets retrieved from Kili API."""
+
+import json
+from typing import Dict
+
+ASSET_JSON_FIELDS = ["jsonMetadata", "labels.jsonResponse", "latestLabel.jsonResponse"]
+
+
+def load_asset_json_fields(asset: Dict, fields) -> Dict:
```
PierreLeveau, line 9: missing types for fields

PierreLeveau (approved):

src/kili/core/helpers.py
Jonas1312: > double underscore prevent the function to be imported in another file. It is to avoid this function to be reused somewhere else.

you can still import it: https://stackoverflow.com/questions/14168791/actual-implementation-of-private-variables-in-python-class

but since we plan to remove it anyway, we can keep it like that 👍

Jonas1312: > I don't make a PR to remove this format result. it is still used in many queries and mutations. We will be able to simply delete it once we will have moved every function. We just have to make sure not to use it anymore

yes what I meant in my comment is that we shouldn't use it for the "new" clean archi

src/kili/adapters/kili_api_gateway/asset/formatters.py
```diff
@@ -0,0 +1,21 @@
...
+    """Load json fields of an asset."""
+    if "jsonMetadata" in asset:
+        asset["jsonMetadata"] = json.loads(asset["jsonMetadata"])
+    if "labels.jsonResponse" in fields:
+        asset_labels = asset.get("labels", [])
+        for label in asset_labels:
+            label["jsonResponse"] = json.loads(label["jsonResponse"])
+    if "latestLabel.jsonResponse" in fields and asset.get("latestLabel") is not None:
+        asset["latestLabel"]["jsonResponse"] = json.loads(asset["latestLabel"]["jsonResponse"])
+    return asset
```
Jonas1312, line 21: I would be super careful here, some fields might actually be None when they are not mandatory in the graphql schema, even though the default value is different.

For example, jsonMetadata is not mandatory, so it might be None, but in theory, we should fetch "{}" instead.

But I trust the schema better than the backend behavior, so I would still use some `.get()` for the dictionaries, and check that the value is a string before putting it in the json.loads

src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -35,8 +36,9 @@ class PaginatedGraphQLQuery:
     It factorizes code for executing paginated queries.
     """
 
-    def __init__(self, graphql_client: GraphQLClient):
+    def __init__(self, graphql_client: GraphQLClient, http_client: HttpClient):
         self._graphql_client = graphql_client
+        self._http_client = http_client
```
Jonas1312, line 41: why do we need the http client?

src/kili/core/utils/pagination.py
```diff
@@ -115,3 +116,14 @@ def generate_variables(batched_properties):
...
+
+def batcher(iterable: Iterator, batch_size: int):
+    """Break iterable into sub-iterables with batch_size elements each.
+
+    The last yielded list will have fewer than n elements if the
+    length of iterable is not divisible by batch_size:
+    """
+    iterator = iter(iterable)
+    while batch := list(islice(iterator, batch_size)):
+        yield batch
```
Jonas1312, line 129: we already have a BatchIteratorBuilder in core/utils/pagination

src/kili/services/copy_project/__init__.py
```diff
@@ -173,17 +175,15 @@ def _copy_assets(self, from_project_id: str, new_project_id: str) -> None:
...
-                downloaded_assets = self._download_assets(from_project_id, fields, tmp_dir, assets)
-                return self._upload_assets(new_project_id, downloaded_assets)
+        assets_gen = self.kili.kili_api_gateway.list_assets(filters, fields, options)
 
-        asset_gen = self.kili.kili_api_gateway.list_assets(
-            filters, fields, options, download_and_upload_assets
-        )
-        # Generator needs to be iterated over to actually fetch assets
-        for _ in asset_gen:
-            pass
```
Jonas1312, line 186: make sure that copy project e2e tests pass, I remember we had issues with this

Jonas1312 (changes requested):

src/kili/adapters/kili_api_gateway/asset/formatters.py
theodu: Yeah, safer indeed

src/kili/adapters/kili_api_gateway/helpers/queries.py
theodu: okay, so we indeed don't need it currently. 
I added it because I used it and that I didn't delete it in case we need it in the future, not to change plenty of files. If you prefer to delete it I can, I don't have a strong opinion on that

Jonas1312: I would not keep it. I don't think we would need it in the future in this part of the code, since we just query list/dict data here. We're not supposed to preprocess/parse it

src/kili/core/utils/pagination.py
theodu: It only work for list because of the __len__ function.
This one is a simpler one and works for generators

src/kili/services/copy_project/__init__.py
theodu: j'ai lancé les tests E2E manuellement

src/kili/adapters/kili_api_gateway/asset/formatters.py
```diff
@@ -0,0 +1,24 @@
...
+
+def load_asset_json_fields(asset: Dict, fields: ListOrTuple[str]) -> Dict:
+    """Load json fields of an asset."""
+    if "jsonMetadata" in fields:
+        asset["jsonMetadata"] = json.loads(asset.get("jsonMetadata", "{}"))
+    if "labels.jsonResponse" in fields:
+        asset_labels = asset.get("labels", [])
+        for label in asset_labels:
+            if (label.get("jsonResponse")) is not None:
+                label["jsonResponse"] = json.loads(label["jsonResponse"])
```
Jonas1312, line 19: here, both labels and jsonResponse are mandatory fields, so no need to check
```suggestion
        asset_labels = asset.get("labels", [])
        for label in asset_labels:
            label["jsonResponse"] = json.loads(label["jsonResponse"])
```

src/kili/adapters/kili_api_gateway/asset/formatters.py
```diff
@@ -0,0 +1,24 @@
...
+    if "latestLabel.jsonResponse" in fields and asset.get("latestLabel") is not None:
+        latest_label = asset["latestLabel"]
+        if (latest_label.get("jsonResponse")) is not None:
+            latest_label["jsonResponse"] = json.loads(latest_label["json_response"])
```
Jonas1312, line 23: jsonresponse is a mandatory field, but yes we need to check for latestLabel 

```suggestion
        asset["latestLabel"]["jsonResponse"] = json.loads(asset["latestLabel"]["json_response"])
```

src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -45,22 +34,20 @@ def execute_query_from_paginated_call(
         where: Dict[str, Any],
         options: QueryOptions,
         tqdm_desc: str,
-        nb_elements_to_query: Optional[int],
-        post_call_function: Optional[Callable[[List], List]] = None,
+        count_query: Optional[str],
```
Jonas1312, line 37: I would keep nb_elements_to_query

not all queries have a count resolver (dataConnections does not have countdataconnections)

Jonas1312: but as discussed, I'm ok to keep self.get_number_of_elements_to_query in the class, totally makes sense, but we can't use it for all queries

src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -102,36 +87,36 @@ def execute_query_from_paginated_call(
                     if len(elements) < first:
                         break
 
+    def get_number_of_elements_to_query(
+        self,
+        count_query: Optional[str],
```
Jonas1312, line 92: I find it weird to make it optional

if we call this method, we must provide the query (if it exists), else it makes no sense to call it

src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -102,36 +87,36 @@ def execute_query_from_paginated_call(
...
+        where: Dict[str, Any],
+        options: QueryOptions,
+    ) -> Optional[int]:
```
Jonas1312, line 95: should always return an int

Jonas1312 (changes requested):

src/kili/services/copy_project/__init__.py
theodu: https://github.com/kili-technology/kili-python-sdk/actions/runs/6110561997

src/kili/adapters/kili_api_gateway/helpers/queries.py
theodu: I agree, changed

src/kili/adapters/kili_api_gateway/helpers/queries.py
theodu: If I do a method to call before to call the execute_query_from_paginated_call, we might forget to call it.
Here it is a it was before in the previous code, if we provide a count_query if it exists, it retrieve the number of elements to call, if not, it will disable tqdm. 
For methods without a count query, we just pass a count_query set as None

Jonas1312: ok 👍

src/kili/adapters/kili_api_gateway/helpers/queries.py
Jonas1312: it's still optional?

src/kili/adapters/kili_api_gateway/helpers/queries.py
Jonas1312: same here

src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -102,36 +87,36 @@ def execute_query_from_paginated_call(
...
-    nb_elements = count_result["data"]
-    nb_elements_queried = max(nb_elements - skip, 0)
-    if first is None:
-        return nb_elements_queried
-    return min(nb_elements_queried, first)
+        Returns:
+            The number of elements to query
+        """
+        if count_query is None:
+            return None
```
Jonas1312, line 110: not needed

Jonas1312 (changes requested):

Jonas1312 (approved):

PR #1441, opened 2023-09-06
feat: add `kili.update_tag`

Jonas1312:
I created a custom type for  List[str] and Tuple[str, ...]  since Sequence[str] also accepts "raw" str

We should type `fields` using this custom type so that:
- we can accept tuples as well as lists
- we can set immutable default values as tuples


PierreLeveau (approved):

src/kili/presentation/client/tag.py
```diff
@@ -129,7 +129,7 @@ def untag_project(
             )
         ]
 
-    def update_tag(self, tag_name: str, new_tag_name: str) -> None:
+    def update_tag(self, tag_name: str, new_tag_name: str) -> str:
```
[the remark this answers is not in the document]
Jonas1312: yes I agree

src/kili/presentation/client/tag.py
```diff
@@ -64,12 +65,16 @@ def tag_project(
         if tag_ids is None:
             if tags is None:
                 raise ValueError("Either `tags` or `tag_ids` must be provided.")
-            tag_ids = tag_use_cases.get_tag_ids_from_labels(labels=tags)
+            tag_ids = tag_use_cases.get_tag_ids_from_labels(
```
theodu, line 68: this should be part of the `tag_project` use case or if it will be used in several useCases. It is rather a gateway operation. a useCase = one client action

theodu: The useCase can take a tags argument that is a dataclass that either have a tag_id field or a tag_label field.
We have a similar issue with assets if we want to switch from a "list of attributes" approach to a "list of dict with attributes" one: how to assert that an asset can be well identified if its id or external ids are optional but that at least one should be provided ?
I began to think about it but didn't do any implem because we didn't move asset mutations yet. 
It can maybe done with Pydantic validators ?

src/kili/use_cases/tag/__init__.py
```diff
@@ -95,3 +97,18 @@ def get_tag_ids_from_labels(self, labels: Sequence[str]) -> List[TagId]:
...
+    ) -> UpdateTagReturnData:
+        """Update tag.
+
+        This operation is organization-wide.
+        The tag will be updated for all projects of the organization.
+        """
+        if tag_id is None:
+            if tag_name is None:
+                raise ValueError("Either `tag_name` or `tag_id` must be provided.")
+            tag_id = self.get_tag_ids_from_labels(labels=(tag_name,))[0]
```
theodu, line 112: see: here use cases are calling each others

theodu (changes requested):

src/kili/use_cases/tag/__init__.py
Jonas1312: yep this one is a bad copy paste, I fixed it

src/kili/presentation/client/tag.py
Jonas1312: > this should be part of the `tag_project` use case or if it will be used in several useCases. It is rather a gateway operation. a useCase = one client action

I asked Marc and it's fine to call multiple use cases sequentially in the presenter. 

Making a big use case makes things harder to read/maintain/test IMO

If you want to put that in a gateway operation, you're going to end up with the gateway calling itself (self.list_tags_by_org), which is bad. A layer method should not call another layer method (from what I understand of clean archi...)

PR #1444, opened 2023-09-06
feat: add `kili.delete_tag()`

Jonas1312:


src/kili/presentation/client/tag.py
```diff
@@ -149,3 +149,25 @@ def update_tag(self, tag_name: str, new_tag_name: str) -> Dict[Literal["id"], st
                 tag_use_cases.update_tag(tag_id=tag_id, new_tag_name=new_tag_name).updated_tag_id
             )
         }
+
+    def remove_tag(self, tag_name: Optional[str] = None, tag_id: Optional[str] = None) -> bool:
```
theodu, line 153: why remove and not delete ? I thought at first that it would remove a tag from a project

theodu (approved):

PR #1445, opened 2023-09-07
feat: `kili.projects()` with `tags_in` filter

Jonas1312:
TODO:

- [x] merge main when #1439 is merged
- [x] make sure e2e tests pass before merging:


src/kili/domain/project.py
```diff
@@ -9,3 +13,20 @@
...
+
+@dataclass
+# pylint: disable=too-many-instance-attributes
+class ProjectFilters:
+    """Project filters for running a project search."""
+
+    id: Optional[ProjectId]  # pylint: disable=invalid-name
+    archived: Optional[bool]
+    search_query: Optional[str]
+    should_relaunch_kpi_computation: Optional[bool]
```
theodu, line 26: wow we still have this field :O. It is really archaic and not useful anymore because we know automatically computes the consensus and honeypot after each label but before we had to send a compute request to update it.
Just for a reminder, We should deprecate it in the presentation method

theodu (approved):

PierreLeveau (approved):

PR #1449, opened 2023-09-08
fix: unnormalized export for PDF projects with NER jobs 

Jonas1312:


src/kili/services/export/format/kili/__init__.py
```diff
@@ -151,10 +151,17 @@ def _scale_json_response_vertices(
         self, json_resp: Dict, asset: Dict, annotation_scaler: Callable[[Dict, Dict], None]
     ) -> None:
         for job_name in json_resp.keys():
-            if (
+            if not (
                 # some old labels might not up to date with the json interface
-                job_name not in self.project["jsonInterface"]["jobs"]
-                or self.project["jsonInterface"]["jobs"][job_name]["mlTask"] != "OBJECT_DETECTION"
+                job_name in self.project["jsonInterface"]["jobs"]
```
PierreLeveau, line 156: worth it to isolate it in a private method for better readability

PierreLeveau (changes requested):

theodu (approved):

PR #1468, opened 2023-09-14
refactor: move the external id to asset id utils

PierreLeveau:


src/kili/use_cases/utils/use_cases_utils.py
```diff
@@ -0,0 +1,82 @@
+"""Utils for use cases."""
```
Jonas1312, line 1: not sure if it's necessary to create a folder "utils" for just one .py

src/kili/use_cases/utils/use_cases_utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(self, kili_api_gateway: KiliAPIGateway) -> None:
+        """Init."""
+        self.kili_api_gateway = kili_api_gateway
+
+    def get_asset_ids_or_throw_error(
+        self,
+        asset_ids: Optional[List[str]],
+        external_ids: Optional[List[str]],
+        project_id: Optional[str],
```
Jonas1312, line 24: should be ListOrTuple and ProjectId types

src/kili/use_cases/utils/use_cases_utils.py
```diff
@@ -0,0 +1,82 @@
...
+    ) -> List[str]:
```
Jonas1312, line 25: should return AssetId

src/kili/use_cases/utils/use_cases_utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+        if asset_ids is None and external_ids is not None and project_id is not None:
+            id_map = self.infer_ids_from_external_ids(external_ids, project_id)
+            asset_ids = [id_map[ext_id] for ext_id in external_ids]
+
+        return asset_ids  # type: ignore
+
+    def infer_ids_from_external_ids(
+        self, asset_external_ids: List[str], project_id: str
+    ) -> Dict[str, str]:
```
Jonas1312, line 37: should map AssetExternalId  to AssetId (types)

Jonas1312 (changes requested):

src/kili/use_cases/utils/use_cases_utils.py
```diff
@@ -0,0 +1,82 @@
...
+from kili.domain.asset.helpers import check_asset_identifier_arguments
+from kili.exceptions import NotFound
+
+
+class UseCasesUtils:
+    """Utils for use cases."""
+
+    def __init__(self, kili_api_gateway: KiliAPIGateway) -> None:
+        """Init."""
+        self.kili_api_gateway = kili_api_gateway
```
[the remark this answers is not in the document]
PierreLeveau: I agree if we do a second use case utils. For now I'll put it into a utils.py as suggested

src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -279,12 +279,12 @@ def update_properties_in_assets(
             )
             raise MissingArgumentError("Please provide either `asset_ids` or `external_ids`.")
 
-        asset_ids = get_asset_ids_or_throw_error(
-            self.kili_api_gateway, asset_ids, external_ids, project_id
+        resolved_asset_ids = UseCasesUtils(self.kili_api_gateway).get_asset_ids_or_throw_error(
+            asset_ids, external_ids, project_id  # type: ignore
```
Jonas1312, line 283: this will ignore type errors for all three arguments, which makes pyright useless

it's best to use typing.cast for arrays, or ProjectId(project_id) for single entities

src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -279,12 +279,12 @@ def update_properties_in_assets(
...
         )
 
         properties_to_batch = process_update_properties_in_assets_parameters(
-            asset_ids,
+            resolved_asset_ids,  # type: ignore
```
Jonas1312, line 287: better cast all type ignores below

Jonas1312 (changes requested):
we shouldn't abuse `# type: ignore`. It makes both defining custom types and using pyright useless

src/kili/entrypoints/mutations/label/__init__.py
```diff
@@ -159,17 +161,22 @@ def append_to_labels(
             user = self.get_user()  # type: ignore  # pylint: disable=no-member
             author_id = user["id"]
 
+        # fmt: off
```
Jonas1312, line 164: why is it needed?

Jonas1312 (approved):

src/kili/use_cases/utils.py
```diff
@@ -0,0 +1,91 @@
...
+from kili.adapters.kili_api_gateway.helpers.queries import QueryOptions
+from kili.core.utils import pagination
+from kili.domain.asset import AssetExternalId, AssetFilters, AssetId
+from kili.domain.asset.helpers import check_asset_identifier_arguments
+from kili.domain.project import ProjectId
+from kili.domain.types import ListOrTuple
+from kili.exceptions import NotFound
+
+
+class UseCasesUtils:
```
theodu, line 15: We plan to do neither a use-case nor a gateway fonction, but a utility function. Here is a Use Case and it is too generic, I am afraid that other dev might add too many use case here as a "trash use case".
I would rather transform it into a `get_asset_ids_or_throw_error` and `infer_ids_from_external_ids` functions that takes `kili_api_gateway` as argument to truely have a utility function and not a utility use case

theodu: and this utility could be in a module in the asset use case, not to have a module with every utility functions for all use cases

theodu (changes requested):

src/kili/use_cases/utils.py
PierreLeveau: no, this is not a use case, but the utils that can be used in the use cases. Can be functions as well, but it feels natural to declare the dependency in the class argument. Can be moved to an `AssetUseCasesUtils` eventually.

src/kili/entrypoints/mutations/label/__init__.py
PierreLeveau: oops not needed anymore, it was not playing well with the type: ignore at the time. Fixing it

src/kili/use_cases/utils.py
theodu: I think that it will be cleaner in AssetUseCasesUtils

PierreLeveau: done

PR #1476, opened 2023-09-21
refactor: create questions new archi

PierreLeveau:


src/kili/adapters/kili_api_gateway/issue/__init__.py
```diff
@@ -18,11 +18,11 @@ class IssueOperationMixin(BaseOperationMixin):
     """GraphQL Mixin extending GraphQL Gateway class with Issue related operations."""
 
     def create_issues(
-        self, type_: IssueType, issues: List[IssueToCreateKiliAPIGatewayInput]
+        self, type_: IssueType, issues: List[IssueToCreateKiliAPIGatewayInput], description: str
     ) -> List[IssueId]:
         """Send a GraphQL request calling createIssues resolver."""
         created_issue_entities: List[IssueId] = []
-        with tqdm.tqdm(total=len(issues), desc="Creating issues") as pbar:
+        with tqdm.tqdm(total=len(issues), desc=description) as pbar:
```
Jonas1312, line 25: I think it would be nice here to be able to disable the pbar if needed

so we should implement what we agreed on few days ago: a ProgressBarOptions dataclass with a disable_tqdm and a description

src/kili/presentation/client/question.py
```diff
@@ -0,0 +1,50 @@
...
+class QuestionClientMethods(BaseClientMethods):
+    """Client presentation methods for questions."""
+
+    @typechecked
+    def create_questions(
+        self,
+        project_id: str,
+        text_array: List[Optional[str]],
+        asset_id_array: Optional[List[str]] = None,
+        asset_external_id_array: Optional[List[str]] = None,
```
Jonas1312, line 23: should accept list or tuples as inputs (str is correct)

src/kili/presentation/client/question.py
```diff
@@ -0,0 +1,50 @@
...
+    ) -> List[Dict]:
```
Jonas1312, line 24: dict (literal "id", str)

src/kili/use_cases/question/question_use_case.py
```diff
@@ -0,0 +1,61 @@
...
+from kili.use_cases.utils import UseCasesUtils
+
+
+@dataclass
+class QuestionToCreateUseCaseInput:
+    """Question to create use case input."""
+
+    text: Optional[str]
+    asset_id: Optional[str]
+    asset_external_id: Optional[str]
```
Jonas1312, line 20: custom types AssetId, AssetExternalId

src/kili/use_cases/question/question_use_case.py
```diff
@@ -0,0 +1,61 @@
...
+    ) -> List[QuestionId]:
+        """Create questions."""
+        if questions[0].asset_id is not None:
+            # we assume that if 1 question asset Id is not None, all there others are too
+            asset_id_array = [AssetId(question.asset_id) for question in questions]  # type: ignore
+            external_id_array = None
+        else:
+            asset_id_array = None
+            external_id_array = [
+                AssetExternalId(question.asset_external_id)  # type: ignore
```
Jonas1312, line 39: better doing runtime checks with to make sure it's really the case

and we should avoid type ignore

src/kili/use_cases/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    ) -> Dict[AssetExternalId, AssetId]:
+        # we batch the queries because too many assets in a "in" query makes the query fail
+        assets_generators = (
+            self.kili_api_gateway.list_assets(
+                AssetFilters(project_id, external_id_strictly_in=external_ids_batch),
+                ["id", "externalId"],
+                QueryOptions(disable_tqdm=True),
+            )
+            for external_ids_batch in pagination.BatchIteratorBuilder(
+                list(asset_external_ids), 1000
```
Jonas1312, line 83: the batch builder should accept list or tuple, we shouldn't need an extra conversion step

Jonas1312 (changes requested):

src/kili/adapters/kili_api_gateway/issue/__init__.py
PierreLeveau: added in a separate ticket

src/kili/use_cases/question/question_use_case.py
PierreLeveau: Iterating one more time over possibly 25 000 (and soon 100.000) assets to find out if the ids are defined or not is costly. If there are mixed values asset_ids and external_ids, then the server will end up throwing an error so the DX should be acceptable. Added the casting.

src/kili/use_cases/base.py
```diff
@@ -0,0 +1,12 @@
+"""Base class for all use cases."""
+
+from kili.adapters.kili_api_gateway import KiliAPIGateway
+
+
+class BaseUseCases:
+    """Base class for all use cases."""
+
+    kili_api_gateway: KiliAPIGateway
```
Jonas1312, line 9: no need to declare this class attribute

the class can be made abstract maybe, since it should not be instantiated directly

also, existing use cases (project, tags, etc.) should inherit from this class

src/kili/use_cases/question/question_use_case.py
```diff
@@ -0,0 +1,60 @@
...
+from kili.use_cases.utils import UseCasesUtils
+
+
+@dataclass
+class QuestionToCreateUseCaseInput:
+    """Question to create use case input."""
+
+    text: Optional[str]
+    asset_id: Optional[AssetId]
+    asset_external_id: Optional[AssetExternalId]
```
Jonas1312, line 20: I think @theodu put that kind of dataclass in a types.py module (cf Issues use cases)

src/kili/use_cases/question/question_use_case.py
```diff
@@ -0,0 +1,60 @@
...
+        """Create questions."""
+        if questions[0].asset_id is not None:
+            # we assume that if 1 question asset Id is not None, all there others are too
+            asset_id_array = [cast(AssetId, question.asset_id) for question in questions]
+            external_id_array = None
+        else:
+            asset_id_array = None
+            external_id_array = [
+                cast(AssetExternalId, question.asset_external_id) for question in questions
+            ]
```
Jonas1312, line 40: no need for neither casts nor `type ignore` here I think



what I meant in my previous comment is thatwe can just check the inputs: `assert all(q.asset_id for q in questions)`

 we don't need to make any call to the backend

src/kili/use_cases/utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class UseCasesUtils:
+    """Utils for use cases."""
+
+    def __init__(self, kili_api_gateway: KiliAPIGateway) -> None:
+        """Init."""
+        self.kili_api_gateway = kili_api_gateway
+
+    def get_asset_ids_or_throw_error(
```
Jonas1312, line 22: this method (and maybe others?) are duplicated in the codebase:

- src/kili/use_cases/utils.py
- src/kili/use_cases/asset/utils.py

src/kili/use_cases/utils.py
```diff
@@ -0,0 +1,89 @@
...
+        """Check if external id to internal id conversion is valid and needed."""
+        check_asset_identifier_arguments(project_id, asset_ids, external_ids)
+
+        if asset_ids is None and external_ids is not None and project_id is not None:
+            id_map = self.infer_ids_from_external_ids(external_ids, project_id)
+            resolved_asset_ids = [id_map[ext_id] for ext_id in external_ids]
+        else:
+            resolved_asset_ids = asset_ids or []
+
+        return cast(List[AssetId], resolved_asset_ids)
```
Jonas1312, line 37: the cast at the end should not be needed (and is actually wrong)

if asset_ids input parameter is a Tuple of AssetIds, then resolved_asset_ids will be a Tuple, the return type too

so the return type can be a ListOrTuple too

Jonas1312 (changes requested):

src/kili/use_cases/question/question_use_case.py
PierreLeveau: since it is always imported with the question use case, no need to put it into a separate file

Jonas1312 (approved):

PR #1504, opened 2023-10-02
build: moved build info to pyproject.toml

PierreLeveau:


pyproject.toml
```diff
@@ -1,3 +1,96 @@
...
+  "mkdocs-material",
+  "mkdocstrings",
+  "mkdocstrings-python-legacy",
+  "mkdocs-click",
+  "mike",
+  "pymdown-extensions",
+  # linting
+  "black",
+  "pre-commit >= 3.3.0, < 4.0.0",
+  "pylint == 2.17.7",
```
Jonas1312, line 79: we can relax pylint version no? they follow sem ver

Jonas1312 (approved):

pyproject.toml
PierreLeveau: there is sem ver, but sometimes there are new verifications in minor changes, which makes the CI break without changing our code.

PR #1524, opened 2023-10-05
test: use kili_admin for tests that require org admin priviledges

PierreLeveau:
e2e: https://github.com/kili-technology/kili-python-sdk/actions/runs/6420949015/job/17434205808


theodu (approved):

tests/e2e/test_tags.py
Jonas1312, line 1: where has this test been moved to?

Jonas1312 (changes requested):
instead of moving all tests, we can create two kili client fixtures `kili` and `kili_admin` in tests/e2e/conftest.py

the kili_admin client use the kili_api_key_admin, and the normal kili client uses the kili_api_key_user

by default, all e2e tests use the `kili` fixture (it's already the case)

and if some e2e tests require admin rights, we pass the kili_admin client

Jonas1312 (approved):

Jonas1312:
do the e2e tests with exports fail when using the kili user client? (they should)

PierreLeveau:
> do the e2e tests with exports fail when using the kili user client? (they should)

there is no export e2e test

Jonas1312:
> > do the e2e tests with exports fail when using the kili user client? (they should)
> 
> there is no export e2e test

notebooks

PR #1545, opened 2023-10-17
fix: use externalId for filename in export when asset content is an url

Jonas1312:


Jonas1312:
> wondering if this logic shouldn't be factored. Should it be applied to the yolo format too?

yolo uses this: `filename = asset["externalId"]`

The whole export service needs a refac :/

PR #1566, opened 2023-10-27
refactor: project queries

Jonas1312:


src/kili/services/plugins/upload.py
```diff
@@ -73,7 +76,7 @@ def check_file_is_txt(path: Path, verbose: bool = True) -> bool:
 
 def check_file_contains_handler(path: Path) -> Tuple[bool, Optional[List[str]]]:
     """Return true if the file contain PluginHandler Class."""
-    with open(path, encoding="utf-8") as file:
+    with Path(path).open(encoding="utf-8") as file:
```
theodu, line 79: isn't it supposed to already be a path ?

PR #1576, opened 2023-11-06
fix: implicit conversion from video label annotations to json response

Jonas1312:


src/kili/adapters/kili_api_gateway/label/operations.py
```diff
@@ -1,6 +1,39 @@
...
+
+from kili.adapters.kili_api_gateway.helpers.queries import (
+    fragment_builder,
+)
+from kili.core.graphql.graphql_client import GraphQLClient
+from kili.domain.label import LabelId
+from kili.domain.types import ListOrTuple
+
+
+def list_annotations(
```
PierreLeveau, line 14: it is strange to put that function here, as in general only mutations and queries are listed. If this is because it is used in several places of the gateway, this deserves to be in a `label/common.py` or `label/utils.py` file.

PierreLeveau: this can also be inside a private method of the `kili_api_gateway`

src/kili/adapters/kili_api_gateway/mixin.py
```diff
@@ -0,0 +1,38 @@
+"""Kili API Gateway module for interacting with Kili."""
```
PierreLeveau, line 1: it is not a mixin. It can be named `kili_api_gateway.py`, then `KiliApiGateway` can be imported in the `__init__.py`.

src/kili/utils/typed_dict.py
```diff
@@ -0,0 +1,18 @@
+"""Utils for TypedDicts."""
```
PierreLeveau, line 1: `utils` should be utilities exposed to the users. This can go to `domain.types`

PierreLeveau (changes requested):

src/kili/adapters/kili_api_gateway/mixin.py
Jonas1312: better not put the class KiliApiGateway in the `__init__`, else, everytime we want to get a specific class or method from the gateway, it will go through the `__init__`, which will import all sub-packages, which causes circular imports issues

src/kili/adapters/kili_api_gateway/label/operations.py
Jonas1312: I put it in common.py

PierreLeveau (approved):

src/kili/adapters/kili_api_gateway/label/annotation_to_json_response.py
```diff
@@ -389,36 +385,15 @@ def _video_object_detection_annotation_to_json_response(
...
+    object_type: str,
+    object_initial_state: List[List[List[Vertice]]],
     initial_state_frame_index: int,
-    object_final_state: List[Vertice],
+    object_final_state: List[List[List[Vertice]]],
     final_state_frame_index: int,
     at_frame: int,
-) -> List[Vertice]:
+) -> List[List[List[Vertice]]]:
     """Interpolate the bounding boxes between two frames."""
```
PierreLeveau, line 419: not only bounding boxes

src/kili/adapters/kili_api_gateway/label/annotation_to_json_response.py
```diff
@@ -427,13 +402,266 @@ def _interpolate_object_(
...
+
+def interpolate_rectangle(
+    *,
+    previous_vertices: List[Vertice],
+    next_vertices: List[Vertice],
+    width: int,
+    height: int,
+    weight: float,
+) -> List[Vertice]:
+    """Interpolate a rectangle."""
```
PierreLeveau, line 643: For clarity, explain the principles behind this operation in the docstring.

src/kili/adapters/kili_api_gateway/label/annotation_to_json_response.py
```diff
@@ -389,36 +385,15 @@ def _video_object_detection_annotation_to_json_response(
...
-    object_initial_state: List[Vertice],
+    object_type: str,
+    object_initial_state: List[List[List[Vertice]]],
     initial_state_frame_index: int,
-    object_final_state: List[Vertice],
+    object_final_state: List[List[List[Vertice]]],
     final_state_frame_index: int,
     at_frame: int,
-) -> List[Vertice]:
+) -> List[List[List[Vertice]]]:
```
PierreLeveau, line 418: For better readability:

- make the functions private
- organize the code like a table of content: first the top level, then the functions called by the top level, then apply the recursion.

src/kili/adapters/kili_api_gateway/label/annotation_to_json_response.py
```diff
@@ -427,13 +402,266 @@ def _interpolate_object_(
...
+                        / (final_state_frame_index - initial_state_frame_index),
+                    )
+                ]
+            ]
+        ]
+
+    if object_type == JobTool.RECTANGLE:
+        return [
+            [
+                interpolate_rectangle(
```
PierreLeveau, line 445: given the complexity, this may deserve a few tests

PierreLeveau (changes requested):

src/kili/adapters/kili_api_gateway/label/annotation_to_json_response.py
```diff
@@ -0,0 +1,696 @@
...
+
+    if object_type == JobTool.RECTANGLE:
+        return [
+            [
+                _interpolate_rectangle(
+                    previous_vertices=object_initial_state[0][0],
+                    next_vertices=object_final_state[0][0],
+                    width=1,
+                    height=1,
+                    weight=(at_frame - initial_state_frame_index)
```
Jonas1312, line 450: @baptiste-olivier I'm not sure about setting this to 1. Do we really need to fetch the assets' resolutions too for each label?

src/kili/services/copy_project/__init__.py
baptiste-olivier, line 1: We may need to copy annotations here when we copy a label.

There is a property that may help to do that. In the `appendManyLabels`:  `previousLabelID`.
When appending a label with a `previousLabelID`, backend will automatically copy all annotations of the previousLabel.

baptiste-olivier: Or you can directly copy all annotations and annotationValues

tests/unit/adapters/kili_api_gateway/label/test_data/test_case_1.py
```diff
@@ -0,0 +1,275 @@
...
+                "isPrediction": False,
+            },
+        ],
+        "path": [],
+    },
+]
+
+expected_json_resp = {
+    "0": {
+        "ANNOTATION_JOB_COUNTER": {"BBOX_JOB": {"TRAIN": 1, "CAR": 1}},
```
baptiste-olivier, line 130: we can ignore this field, as it is used only in frontend logic

src/kili/domain/annotation.py
```diff
@@ -0,0 +1,126 @@
...
+    isPrediction: bool
+    frame: int
+    annotationValue: ClassificationAnnotationValue
+
+
+class VideoTranscriptionKeyAnnotation(TypedDict):
+    """Video transcription key annotation."""
+
+    id: KeyAnnotationId
+    isPrediction: bool
```
baptiste-olivier, line 75: Care, as `isPrediction` is moving to annotationValue : https://gitlab.com/kili-technology/kili/-/merge_requests/11110

src/kili/services/copy_project/__init__.py
Jonas1312: we'll cover the copy project case in another PR

PierreLeveau (approved):

baptiste-olivier (approved):

PR #1598, opened 2023-11-28
fix: fix export when an asset has a slash

romainrousseaux:
for example, `as/set.png` will be exported as : 

- labels/
  - as /
    - set.png

### DIFF
--- a/src/kili/services/export/format/voc/__init__.py
+++ b/src/kili/services/export/format/voc/__init__.py
@@ -123,7 +123,6 @@ def _process_asset(
         )
         xml_filename = f'{asset["externalId"]}.xml'
         filepath = labels_folder / xml_filename
-        print(filepath)
         filepath.parent.mkdir(parents=True, exist_ok=True)
         with open(filepath, "wb") as fout:
             fout.write(f"{annotations}\n".encode())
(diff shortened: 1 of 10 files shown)


src/kili/services/export/format/voc/__init__.py
```diff
@@ -121,7 +122,10 @@ def _process_asset(
             json_response, width, height, parameters, valid_jobs
         )
         xml_filename = f'{asset["externalId"]}.xml'
-        with open(labels_folder / xml_filename, "wb") as fout:
+        filepath = labels_folder / xml_filename
+        print(filepath)
```
Jonas1312, line 126: remove prints (use warnings, or logs if need to write to console)

tests/unit/services/export/test_geojson.py
```diff
@@ -64,3 +64,39 @@ def test_kili_export_labels_geojson(mocker: pytest_mock.MockerFixture):
 
     assert output["type"] == "FeatureCollection"
     assert len(output["features"]) == 5  # 5 annotations in geotiff_image_project_assets.json
+
+
+def test_process_asset_with_external_id_containing_slash():
+    with TemporaryDirectory() as f:
```
Jonas1312, line 70: you can get a temp dir using pytest fixture: https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture

Jonas1312 (commented):
thanks for  the PR

can you fix the pyright error too?

PierreLeveau (approved):

Jonas1312 (approved):

Jonas1312:
wait for e2e tests to pass before merging: https://github.com/kili-technology/kili-python-sdk/actions/runs/7062228310

PR #1601, opened 2023-12-05
fix(LOPS-1098): fix case on create user email. force it to lower

Tako:


PierreLeveau (changes requested):
there are unit tests failing (so in fact this was tested, but with wrong expected values). You can change the expected value to {"data": ... and this should be good.

src/kili/adapters/kili_api_gateway/user/operation_mixin.py
```diff
@@ -53,7 +53,7 @@ def create_user(self, data: CreateUserDataKiliGatewayInput, fields: ListOrTuple[
         """Create a user."""
         fragment = fragment_builder(fields)
         query = get_create_user_mutation(fragment)
-        variables = create_user_data_mapper(data)
+        variables = {"data": create_user_data_mapper(data)}
```
Jonas1312, line 56: we prefer put the data here, just a convention. (see update_password method below)

Jonas1312 (approved):

src/kili/adapters/kili_api_gateway/user/operation_mixin.py
FannyGaudin: Oh ok no problem

PR #1619, opened 2024-01-08
fix(lab-2558): import jp2 and ntf images asynchronously

josselinbuils:
These images require post-processing either because they can contain geospatial metadata, or because they are not directly supported by the browser.


Jonas1312 (approved):

baptiste-olivier (approved):

theodu (approved):

src/kili/services/asset_import/base.py
```diff
@@ -441,6 +442,13 @@ def check_mime_type_compatibility(self, path: str):
         if not os.path.isfile(path):
             raise FileNotFoundError(f"file {path} does not exist")
         mime_type, _ = mimetypes.guess_type(path)
+        if mime_type is None and platform.system() == "Windows":
+            try:
+                import magic
+
+                mime_type = magic.from_file(path, mime=True)
```
josselinbuils, line 449: Not sure if `magic.from_file` returns something else than `None` or an error in case it does not find the mime type but I was not able to find the answer neither in their doc (quite poor) or in the code :(

Jonas1312: yep not sure how we can do this easily...

https://github.com/ahupp/python-magic/pull/294 could help

Jonas1312 (approved):
we can merge once e2e pass: https://github.com/kili-technology/kili-python-sdk/actions/runs/7460684592  (ignore test_cloud_storage flaky test)

PR #1624, opened 2024-01-15
feat(LOPS-1103): fix doc and change naming for consitency

FannyGaudin:


src/kili/use_cases/cloud_storage/__init__.py
```diff
@@ -255,9 +255,10 @@ def update_data_integration(
         status: Optional[DataIntegrationStatus],
     ) -> Dict:
         """Update data integration."""
+        formated_organization_id = organization_id if isinstance(organization_id, str) else ""
```
Jonas1312, line 258: should be :

```python
organization_id=OrganizationId(organization_id) if organization_id else None,
```

not: formated_organization_id = organization_id if isinstance(organization_id, str) else ""

src/kili/presentation/client/cloud_storage.py
```diff
@@ -388,16 +388,14 @@ def create_cloud_storage_integration(
...
+        # pylint: disable=line-too-long
         """Create a cloud storage integration.
 
         Args:
-            fields: All the fields to request among the possible fields for the
-            cloud storage integrations.
-                See [the documentation]
-                (https://docs.kili-technology.com/reference/graphql-api#dataintegration)
-                for all possible fields.
+            fields: All the fields to request among the possible fields for the cloud storage integrations.
```
Jonas1312, line 395: for the cloud storage integration

src/kili/presentation/client/cloud_storage.py
```diff
@@ -499,10 +497,10 @@ def update_data_integration(
         """
         return CloudStorageUseCases(self.kili_api_gateway).update_data_integration(
             name=name,
-            data_integration_id=DataIntegrationId(data_integration_id),
+            cloud_storage_integration_id=DataIntegrationId(cloud_storage_integration_id),
```
Jonas1312, line 500: for use cases, gateway, you can keep the "data integration, data connection"

we only use "cloud storage" for doc and user facing methods

src/kili/use_cases/cloud_storage/__init__.py
FannyGaudin: Yes but if I do it like this i have this error :
`Expression of type "tuple[OrganizationId | None]" cannot be assigned to declared type "str | None"`

src/kili/presentation/client/cloud_storage.py
FannyGaudin: Yes but in this file cloud_storage_integration_id was used everywhere so we decided to use it as well for keep consistency, is it ok for you ?

src/kili/use_cases/cloud_storage/__init__.py
Jonas1312: you  can type the organization_id field of DataIntegrationData  to a Optional[OrganizationId]

FannyGaudin: Oh yes my bad I already did but the formatter didn’t work properly, thanks

src/kili/presentation/client/cloud_storage.py
FannyGaudin: I have update the code for be more clear

src/kili/presentation/client/cloud_storage.py
FannyGaudin: Done

HNicolas (approved):

Jonas1312 (approved):

PR #1633, opened 2024-01-23
fix(LAB-2622): delete project at the beginning if it exist

FannyGaudin:
When testing append_labels, on the backend side there is an async function that compute the KPI of every label, as it take a lot of time if we delete the project at the end of the test set, the compute KPI will fail (it need the project).
So we delete it at the beginning of the test set.


tests/e2e/test_mutations_label.py
```diff
@@ -19,11 +19,18 @@ def project(kili: Kili):
             }
         }
     }
+    project_name = "test_e2e_mutations_labels"
+
+    project_ids = kili.projects(search_query=f"%{project_name}%")
```
baptiste-olivier, line 24: you should specify that you are interested only in field id here

FannyGaudin: yep done

baptiste-olivier (approved):

Jonas1312 (commented):
there are two fixtures (@pytest.fixture()) in this module:
- project
- project_and_asset_id

since the flaky test (test_append_many_labels) uses the fixture project_and_asset_id, shouldn't we modify this one instead?

Jonas1312 (approved):

PR #1634, opened 2024-01-24
feat(LAB-2654): apply points reordering for polygon export

BlueGrizzliBear:
Corresponding ticket : https://linear.app/kili-technology/issue/LAB-2654/aau-when-i-export-my-labels-using-geojson-format-through-the-sdk-my

This MR focuses on : 
- Fixing the export of **polygon** annotations, to always have anti-clockwise order points in GeoJson features
- Adding a unit test, when 2 polygons : one with clockwise points order and one with anti-clockwise points order. Both polygons should be convert to anti-clockwise GeoJson features


src/kili/utils/labels/geojson/polygon.py
```diff
@@ -3,6 +3,35 @@
...
+
+    for i in range(n):
+        x1, y1 = vertices[i]["x"], vertices[i]["y"]
+        x2, y2 = vertices[(i + 1) % n]["x"], vertices[(i + 1) % n]["y"]
+        sum_product += (x2 - x1) * (y2 + y1)
+
+    return sum_product > 0
+
+
+def order_counter_clockwise(vertices):
```
baptiste-olivier, line 22: this function assume that polygons are either drawn in clockwise or counter clockwise order.

The platform support drawing of polygons that are neither clock or counter clockwise ordered.
Exemple : 
not-ordered-polygon

As those polygons maybe not valid for geojson format, we should ignore them to avoid blocking export format, and add a warning to the user.

theodu (approved):

baptiste-olivier (approved):

PR #1638, opened 2024-01-26
test(LAB-2654): allow intersected geojson polygons if with order

BlueGrizzliBear:
Context : 

For crossing edges polygons, instead of not managing these polygons, the geojson export should try to analyse the order of the polygon, and if only the order could not be determined, should return a warning and not deal with the polygon.


baptiste-olivier (approved):

src/kili/utils/labels/geojson/polygon.py
```diff
@@ -7,72 +7,9 @@
...
-    """Returns a boolean to indicate if the first vertex (q) is on the segment (pr)."""
-    qy_not_on_segment = min(py, ry) <= qy <= max(py, ry)
-    qx_not_on_segment = min(px, rx) <= qx <= max(px, rx)
-
-    return qy_not_on_segment and qx_not_on_segment
-
-
-def is_clockwise(vertices):
-    """Returns a boolean to indicate if the vertices are stored clockwise in the array.
+def get_order(vertices):
```
HNicolas, line 10: The name is not really explicit, `get_oriented_area` instead ?
Weird that it returns a positive number for negative polygons, if it is only used here maybe we could fix the formula `(x1 - x2)` and fix the check where the function is called.

BlueGrizzliBear: I think it's only a question of reference.
But, yes, Math-wise, if my polygon is anti-clockwise, it would make more sens if the area value returned is negative 👍 
I'm switching signs, comparison and method name for a more appropriate one.

HNicolas (approved):

PR #1653, opened 2024-02-14
chore(deps): upgrades pyrate-limiter

mjvankampen:


src/kili/core/graphql/graphql_client.py
```diff
@@ -303,7 +303,8 @@ def _execute_with_retries(
     def _raw_execute(
         self, document: DocumentNode, variables: Optional[Dict], **kwargs
     ) -> Dict[str, Any]:
-        with _limiter.ratelimit("GraphQLClient.execute", delay=True), _execute_lock:
+        _limiter.try_acquire("GraphQLClient.execute")
```
Jonas1312, line 306: I think that this instruction does nothing (it's not blocking): https://pyratelimiter.readthedocs.io/en/latest/modules/pyrate_limiter.limiter.html?highlight=try_acquire#pyrate_limiter.limiter.Limiter.try_acquire

The current code with the context manager is blocking

Jonas1312 (changes requested):

src/kili/core/graphql/graphql_client.py
mjvankampen: By setting max_delay it is blocking in the limiter itself (now set to two minutes) it is blocking for max. two minutes.

Ref. https://pyratelimiter.readthedocs.io/en/latest/#rate-limit-delays
and the unit test

PR #1665, opened 2024-03-12
feat(lab-2753): add support for multilayer

HNicolas:
This PR should be merged after https://gitlab.com/kili-technology/kili/-/merge_requests/11663


baptiste-olivier (commented):
Shouldn't we add some test for this new case of import, as made in `tests/unit/services/asset_import/test_import_video.py` ?

baptiste-olivier (approved):

PR #1696, opened 2024-05-22
feat: add issue status update

paul-godhouse:


src/kili/presentation/client/issue.py
```diff
@@ -209,3 +209,16 @@ def issues(
         if as_generator:
             return issues_gen
         return list(issues_gen)
+
+    def update_issue_status(self, issue_id: IssueId, status: IssueStatus) -> Dict[str, Any]:
```
Jonas1312, line 213:
```suggestion
    def update_issue_status(self, issue_id: str, status: IssueStatus) -> Dict[str, Any]:
```

Jonas1312 (commented):
it's not recommended to use NewType custom types for user-facing methods

the SDK user will have type check errors like  this:

```python
from typing import NewType

IssueId = NewType("IssueId", str)


def my_func(issue_id: IssueId) -> None:
    pass


my_func("0123456789")

```

```
Argument of type "Literal['0123456789']" cannot be assigned to parameter "issue_id" of type "IssueId" in function "my_func"
  "Literal['0123456789']" is incompatible with "IssueId"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```

src/kili/presentation/client/issue.py
paul-godhouse: @Jonas1312 I guess it's the same for IssueStatus then?

Jonas1312: not it works fine with Literal types

image

You can render the documentation using `mkdocs serve`, and check that your user-facing method accepts native python types.

The user shouldn't have to import anything from `kili.domain.***`

PR #1713, opened 2024-06-17
feat: add export llm_v1

HNicolas:


src/kili/services/export/format/llm/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+        models = []
+
+    with Path(asset["content"]).open(encoding="utf8") as file:
+        data = json.load(file)
+    version = data.get("version", None)
+    if version == "0.1":
+        for index, prompt in enumerate(data["prompts"]):
+            raw_data.append(
+                {
+                    "role": prompt.get("title", None),
```
BlueGrizzliBear, line 117: There is no key "title" on the prompt object unfortunately.
Maybe we can have a default value of "user" in this case ? 
See discussion on slack.

BlueGrizzliBear (approved):
LGTM 👏

baptiste-olivier (approved):

PR #1732, opened 2024-07-01
fix(LAB-2909): update doc about json_metadata and processingParameters

FannyGaudin:


docs/sdk/tutorials/importing_video_assets.md
```diff
@@ -222,6 +223,39 @@ assets = kili.append_many_to_dataset(
 )
 ```
 
+## Updating video metadata
+
+Except the processingParameters, json metadatas aren't persistent, this means that if you want to update one metadata you will have to pass all the others metadatas to keep them, otherwise it will be erased.
```
stephnguyenkili, line 228: A suggestion to make things less ambiguous:
When importing a video, if you want to add or delete metadata, you must provide the entire list of metadata. This new list will replace the previous one, meaning any metadata not included in the new list will be erased. However, this does not apply to the `processingParameters` metadata, which will be retained even if they are not present in the new list.

FannyGaudin: That's better thanks, i've update the doc

docs/sdk/tutorials/importing_video_assets.md
```diff
@@ -222,6 +223,39 @@ assets = kili.append_many_to_dataset(
 )
 ```
 
+## Updating video metadata
+
+When importing a video, if you want to add or delete metadata, you must provide the entire list of metadata. This new list will replace the previous one, meaning any metadata not included in the new list will be erased. However, this does not apply to the `processingParameters` metadata, which will be retained even if they are not present in the new list.
```
aurelienlombard, line 228: I find this sentence not so clear, especially that after this you provide an example where you update the processing parameters. I would say that they will not be modified, except if the processingParameters are explicitly provided. And maybe explain that they have to do it only if they understand what they do as it can have an impact on how the video will be managed by the application.

aurelienlombard (changes requested):

docs/sdk/tutorials/importing_video_assets.md
FannyGaudin: Yes i've update the documentation, thanks

aurelienlombard (approved):

baptiste-olivier (approved):

PR #1733, opened 2024-07-03
feat(LAB-2960): use existing project to create a new one

FannyGaudin:


aurelienlombard (approved):

src/kili/use_cases/project/project.py
```diff
@@ -28,18 +28,61 @@ def create_project(
...
+                json_interface=project_copied["jsonInterface"],
+                title=title,
+                description=description,
+                project_type=project_type,
+                compliance_tags=compliance_tags,
+            )
+            if project_copied["instructions"]:
+                self.update_properties_in_project(
+                    project_id=new_project_id,
+                    can_navigate_between_assets=None,
```
baptiste-olivier, line 54: Do we need to specify the None parameters, or can we ommit them ?

FannyGaudin: I didn’t manage to ommit it but maybe you have a better solution ?

src/kili/use_cases/project/project.py
```diff
@@ -28,18 +28,61 @@ def create_project(
...
+                    number_of_remaining_assets=None,
+                    number_of_reviewed_assets=None,
+                    review_coverage=None,
+                    should_relaunch_kpi_computation=None,
+                    use_honeypot=None,
+                    metadata_types=None,
+                    title=None,
+                )
+            for tag in project_tag:
+                self._kili_api_gateway.check_tag(project_id=new_project_id, tag_id=tag["id"])
```
baptiste-olivier, line 76: Tags exist from an organization point of view.

If user of organization A is admin of project P of organization B, and try to create a project from P, it may have an error as tags will differ between organizations

tests/integration/use_cases/test_project.py
```diff
@@ -25,6 +26,24 @@ def test_when_create_project_it_works(kili_api_gateway: KiliAPIGateway):
...
+        json_interface={},
+        title="test",
+        description="description",
+        project_id=ProjectId("fake_project_id"),
+        project_type=None,
+        compliance_tags=None,
+    )
+
+    # Then
+    assert project_id == "fake_copied_project_id"
```
baptiste-olivier, line 44: I'm not sure about this test, what are we testing ?

src/kili/use_cases/project/project.py
baptiste-olivier: for my understandig they are all defined as Optional, so you should be able to ignore all Optional one.

What error do you have when you omit them ?

tests/integration/use_cases/test_project.py
FannyGaudin: That the id is not the same as the projectId passed as parameters, but maybe it's useless

src/kili/use_cases/project/project.py
FannyGaudin: If I remove them i have this error
`Arguments missing for parameters "project_id", "can_navigate_between_assets", "can_skip_asset", "compliance_tags", "consensus_mark", "consensus_tot_coverage", "description", "honeypot_mark", "input_type", "json_interface", "min_consensus_size", "number_of_assets", "number_of_skipped_assets", "number_of_remaining_assets", "number_of_reviewed_assets", "review_coverage", "should_relaunch_kpi_computation", "title", "use_honeypot", "metadata_types"`
They are mandatory in the method `update_properties_in_project` in the same file, even if I use `self._kili_api_gateway.update_properties_in_project` I have the same error

src/kili/use_cases/project/project.py
FannyGaudin: Ok I have update the code for use the method from the `TagUseCases` which already handle this case and throw an error if it's the case

baptiste-olivier: so maybe check before if we are on the same organization of the project or if tags exists on your organization.
If it's not the case, we should warn the user that tags hasnt been copied

FannyGaudin: Ok I have update the code accordingly

src/kili/use_cases/project/project.py
baptiste-olivier: Yes, we need to specify default value for parameters to be optional, can we add those ?🙏

FannyGaudin: Yes ok, you’re right that's better, i've update the code

baptiste-olivier (approved):

PR #1735, opened 2024-07-04
fix(lab-2558): fix area property of COCO export and ignore empty bboxes

josselinbuils:
Currently it contains the area of the image instead of the area of the annotation and some annotations are exported with null dimensions.

### DIFF
--- a/src/kili/services/export/format/coco/__init__.py
+++ b/src/kili/services/export/format/coco/__init__.py
@@ -6,6 +6,11 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
+import numpy as np
+from shapely.geometry import Polygon
+from shapely.ops import polygonize
+from shapely.validation import make_valid
+
 from kili.domain.ontology import JobMLTask, JobTool
 from kili.services.export.exceptions import (
     NoCompatibleJobError,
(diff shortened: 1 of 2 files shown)


src/kili/services/export/format/coco/__init__.py
```diff
@@ -429,7 +432,7 @@ def _get_coco_image_annotations(
             # But a type of object can appear several times on the same image.
             # The limitation of the single connected part comes from Kili.
             segmentation=[poly],
-            area=coco_image["height"] * coco_image["width"],
+            area=bbox[2] * bbox[3],
```
aurelienlombard, line 435: I don't think that this area is ok for an annotation that is not a bounding box. It is supposed to be the area of the annotation and not the area of the bounding box surrounding the shape.

aurelienlombard (changes requested):

src/kili/services/export/format/coco/__init__.py
josselinbuils: Hum, you're right, do we you know how to compute easily the area of complex forms?

src/kili/services/export/format/coco/__init__.py
```diff
@@ -442,6 +450,23 @@ def _get_coco_image_annotations(
     return coco_annotations, annotation_j
 
 
+# Shoelace formula, implementation from https://stackoverflow.com/a/30408825.
+def _get_shaleace_area(x: List[float], y: List[float]):
```
HNicolas, line 454: shoelace ?

josselinbuils: yep: https://en.wikipedia.org/wiki/Shoelace_formula

josselinbuils: Fixed ✅

HNicolas (approved):

aurelienlombard (approved):

PR #1738, opened 2024-07-09
feat: add an export format for dynamic llm annotations

baptiste-olivier:


HNicolas (approved):
It is sad to have hardcoded ranking values but I guess you didn't find a way to do this differently.

baptiste-olivier:
> It is sad to have hardcoded ranking values but I guess you didn't find a way to do this differently.

Yes, not a lot of options for now with the current version of dynamic labeling, but will be better for the next one

tests/unit/services/export/test_llm.py
```diff
@@ -0,0 +1,369 @@
...
+                    "id": "id_1",
+                    "chat_id": "global-chat-id",
+                    "model": None,
+                },
+                {
+                    "role": "Model B",
+                    "content": "Hello! How can I assist you today? If you have any questions or need information, feel free to ask. I'm here to help!",
+                    "id": "id_1_2",
+                    "chat_id": "global-chat-id",
+                    "model": None,
```
baptiste-olivier, line 300: Checked with @jeanlatapy , this should be `model-2`

josselinbuils (approved):

PR #1739, opened 2024-07-09
test(LAB-2558): add negative polygon in COCO format and fix/add tests

josselinbuils:


baptiste-olivier (approved):

src/kili/services/export/format/coco/__init__.py
```diff
@@ -478,17 +478,19 @@ def _get_coco_geometry_from_kili_bpoly(
     x_max, y_max = max(p_x), max(p_y)
     bbox_width, bbox_height = x_max - x_min, y_max - y_min
     area = _get_shoelace_area(p_x, p_y)
+    polygons = [[p for vertice in poly_vertices for p in vertice]]
 
     # Compute and remove negative area
     if len(bounding_poly) > 1:
         negative_normalized_vertices = bounding_poly[1]["normalizedVertices"]
```
aurelienlombard, line 485: did not see this in the previous review but you can potentially have multiple holes so you should iterate on the bounding polys

aurelienlombard (changes requested):

src/kili/services/export/format/coco/__init__.py
josselinbuils: Fixed and tested ✅

aurelienlombard (approved):

PR #1743, opened 2024-07-17
fix(LAB-3006): remove duplicated assets when exporting labels

josselinbuils:


src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -57,6 +60,11 @@ def execute_query_from_paginated_call(
         )
         disable_tqdm = nb_elements_to_query is None or options.disable_tqdm
 
+        # If we need an unicity check, it means that new values can be inserted during the
+        # pagination process, so that the initial number of element to query cannot be trusted.
+        use_nb_elements_to_query = unicity_field is None or options.first is not None
+        unicity_values = []
```
baptiste-olivier, line 66: as unicity_values is just here to check the existence or not of the unicity value, it should be a dict to check it in less complexity

src/kili/adapters/kili_api_gateway/helpers/queries.py
```diff
@@ -86,10 +95,19 @@ def execute_query_from_paginated_call(
...
 
-                    yield from elements
+                    if unicity_field is None:
+                        yield from elements
+                        pbar.update(len(elements))
+                    else:
+                        for element in elements:
+                            unicity_value = element[unicity_field]
+
+                            if unicity_value not in unicity_values:
```
baptiste-olivier, line 105: else, shouldn't we increase the `count_elements_retrieved` ?

josselinbuils: In case of unicity check, `count_elements_retrieved` is only used to determine the next elements to retrieve so it should not depend on the duplicates.

When unicity check is enabled, we retrieve elements until there is no more to get if `first` property is not provided, otherwise we stop when the number of unique elements reach `first` value.

aurelienlombard (approved):

baptiste-olivier (approved):

PR #1752, opened 2024-07-29
fix: add try catch for response.json() to catch error on export

FannyGaudin:


src/kili/core/graphql/graphql_client.py
```diff
@@ -224,6 +228,12 @@ def _get_kili_app_version(self) -> Optional[str]:
         """
         url = self.endpoint.replace("/graphql", "/version")
         response = self.http_client.get(url, timeout=30)
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as err:
+            self.logger.warning(f"An error occurred while processing the response: {err}")
+            raise RuntimeError(f"An error occurred while processing the response: {err}")
```
FannyGaudin, line 235: I'm not sure about the RuntimeError maybe there is a better Error ?

src/kili/log/logging.py
```diff
@@ -0,0 +1,7 @@
+import logging
+import sys
+
+logger = logging.getLogger("kili")
```
Jonas1312, line 4: you can use this logger everywhere in the codebase

replace the several `getLogger( kili something, or __name__ )`  with the new global logger

https://github.com/search?q=repo%3Akili-technology%2Fkili-python-sdk%20getLogger&type=code

.github/scripts/upload_test_stats_datadog.py
```diff
@@ -77,6 +79,11 @@ def get_workflow_runs_from_github() -> List[Dict]:
     while True:
         print("Fetching page", page, "...")
         response = requests.get(url + f"&page={page}", headers=HEADERS, timeout=30)
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as err:
+            logger.exception("An error occurred while processing the response: %s", err)
+            raise RuntimeError(f"An error occurred while processing the response: {err}") from err
```
Jonas1312, line 86:
```suggestion
            raise
```

you don't have to raise a custom error, logging + re-raising the requests error can be enough

but I'm fine with the runtimeerror too

src/kili/adapters/kili_api_gateway/asset/formatters.py
```diff
@@ -3,18 +3,27 @@
...
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError as err:
+        logger.exception("An error occurred while processing the response: %s", err)
+        raise RuntimeError(f"An error occurred while processing the response: {err}") from err
     try:
-        return http_client.get(link, timeout=30).json()
+        return response.json()
     except json.JSONDecodeError:
         return {}
```
Jonas1312, line 28: we should log here maybe, if the json can't be parsed (not sure why/when it happens...?)

Jonas1312 (commented):
since the same code is repeated everywhere, we could create a dedicated function log_raise_for_status(requests.Response)

that calls response.raise_for_status() and logs if there's an error

src/kili/log/logging.py
FannyGaudin: yes i'll done it in a second MR

Jonas1312 (approved):

aurelienlombard (approved):

PR #1784, opened 2024-10-01
fix: replace openai dummy endpoint by dummy strings to avoid lychee fail

BlueGrizzliBear:
This MR is to fix the urls present on the code that make the Lychee job fail.
The urls targeted by this MR : the openai urls (not the one regarding the tutorial)

### DIFF
--- a/src/kili/llm/presentation/client/llm.py
+++ b/src/kili/llm/presentation/client/llm.py
@@ -118,7 +118,7 @@ class LlmClientMethods:
             ...     "type": "OPEN_AI_SDK",
             ...     "credentials": {
             ...         "api_key": "your_open_ai_api_key",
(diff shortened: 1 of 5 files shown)


src/kili/llm/presentation/client/llm.py
```diff
@@ -118,7 +118,7 @@ def create_model(self, organization_id: str, model: dict) -> ModelDict:
             ...     "type": "OPEN_AI_SDK",
             ...     "credentials": {
             ...         "api_key": "your_open_ai_api_key",
-            ...         "endpoint": "https://api.openai.com/v1/"
+            ...         "endpoint": "ypur_open_ai_endpoint"
```
RuellePaul, line 121: petite typo

RuellePaul (approved):

HNicolas (approved):

PR #1806, opened 2024-10-29
fix(LAB-3228): fix dogfooding with system prompt export

aurelienlombard:


josselinbuils (approved):

baptiste-olivier:
Would be great to enrich the existing test to add pre-prompt, to be sure that this problem won't appear again

baptiste-olivier (approved):

PR #1812, opened 2024-11-07
feat(LAB-3244): on LLM dynamic projects export annotations at convers…

FannyGaudin:
…ation level

### DIFF
--- a/src/kili/llm/services/export/dynamic.py
+++ b/src/kili/llm/services/export/dynamic.py
@@ -50,6 +50,7 @@ class LLMDynamicExporter:
                 chat_items = label["chatItems"]
                 annotations = label["annotations"]
                 rounds = self._build_rounds(chat_items, annotations, json_interface)
+                total_rounds = len(rounds)
                 for step, round in enumerate(rounds):
                     raw_data = _format_raw_data(
                         round["context"]
@@ -59,25 +60,30 @@ class LLMDynamicExporter:
                         label["id"],
                         obfuscated_models,
                     )
+                    formatted_response = _format_json_response(
+                        json_interface["jobs"],
+                        round["annotations"],
+                        round["completion"],
+                        obfuscated_models,
+                    )
+                    label_data = {
+                        "author": label["author"]["email"],
+                        "created_at": label["createdAt"],
+                        "label_type": label["labelType"],
+                        "label": {},
+                    }
+                    if formatted_response["turn"]:
+                        label_data["label"]["turn"] = formatted_response["turn"]
+                    if step == total_rounds - 1 and formatted_response["conversation"]:
+                        label_data["label"]["conversation"] = formatted_response["conversation"]
+
                     result[f"{step}"] = {
                         "external_id": asset["externalId"],
                         "metadata": asset["jsonMetadata"],
                         "models": _format_models_object(
                             asset["assetProjectModels"], obfuscated_models
                         ),
-                        "labels": [
-                            {
-                                "author": label["author"]["email"],
-                                "created_at": label["createdAt"],
-                                "label_type": label["labelType"],
-                                "label": _format_json_response(
-                                    json_interface["jobs"],
-                                    round["annotations"],
-                                    round["completion"],
-                                    obfuscated_models,
-                                ),
-                            }
-                        ],
+                        "labels": [label_data],
                         "raw_data": raw_data,
                         "status": asset["status"],
                     }
@@ -173,6 +179,9 @@ class LLMDynamicExporter:
                         has_children = True
                         parent_target = chat_item["id"]
 
+                current_round["annotations"] += [
+                    annotation for annotation in annotations if annotation["chatItemId"] is None
+                ]
                 rounds.append(current_round)
                 new_context = (
                     current_round["context"]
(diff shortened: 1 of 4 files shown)


src/kili/llm/services/export/dynamic.py
```diff
@@ -192,8 +195,8 @@ def _format_comparison_annotation(annotation, completions, job, obfuscated_model
 
 def _format_json_response(
     jobs_config: Dict, annotations: List[Dict], completions: List[Dict], obfuscated_models: Dict
-) -> Dict[str, Union[str, List[str]]]:
-    result = {}
+) -> Dict[str, Dict]:
```
HNicolas, line 198: Don't you loose some of the typing here ?

src/kili/llm/services/export/dynamic.py
```diff
@@ -60,25 +61,28 @@ def export(
...
+                        obfuscated_models,
+                    )
+                    label_data = {
+                        "author": label["author"]["email"],
+                        "created_at": label["createdAt"],
+                        "label_type": label["labelType"],
+                        "label": formatted_response["label"],
+                    }
+                    if step == total_rounds - 1 and formatted_response["conversation_label"]:
+                        label_data["conversation_label"] = formatted_response["conversation_label"]
```
baptiste-olivier, line 77: I discussed with Stéphanie about that,  if we want to differentiate the type of the annotation at the label level, we need to do it for every levels ; for example : 

```
...
    "label": {
            "completion": {
                "RANKING": [
                    "B_2"
                ]
            },
            "conversation": {
                "TRANSCRIPTION_1": "some transcription at asset level"
            },
            "chat_item": {
                "012399314932R4": {
                    "TRANSCRIPTION_2": "some transcription at chat item level"
                }
            }
        }
```

baptiste-olivier: I'm not really ok with this format, as we duplicate some information already present in the json_interface. We can discuss it with the team if you want

baptiste-olivier: And doing this change mean applying it also on static export for consistency,  creating a breaking change in our export format

src/kili/llm/services/export/dynamic.py
FannyGaudin: Yes I will update the type

src/kili/llm/services/export/dynamic.py
FannyGaudin: No problem but could you tell me what format you have in mind ?

src/kili/llm/services/export/dynamic.py
```diff
@@ -168,6 +172,7 @@ def _build_rounds(self, chat_items, annotations, json_interface):
                         annotation
                         for annotation in annotations
                         if annotation["chatItemId"] == chat_item["id"]
+                        or annotation["chatItemId"] is None
```
baptiste-olivier, line 175: Not sure we should add this condition here.

If we do so, for each assistant response we will add all annotations without chatItemId.
If our asset haven't any assistant response, annotation whithout chat item id won't be added.

For me we should do it separately, just before appending the current_round to rounds.

src/kili/llm/services/export/dynamic.py
baptiste-olivier, line 1: We should modify static export as well, to keep it in sync with dynamic export

src/kili/llm/services/export/dynamic.py
```diff
@@ -59,25 +60,30 @@ def export(
...
+                    )
+                    label_data = {
+                        "author": label["author"]["email"],
+                        "created_at": label["createdAt"],
+                        "label_type": label["labelType"],
+                        "label": {},
+                    }
+                    if formatted_response["turn"]:
+                        label_data["label"]["turn"] = formatted_response["turn"]
+                    if step == total_rounds - 1 and formatted_response["asset"]:
```
baptiste-olivier, line 77: Did you double check with @stephnguyenkili if we wanted this information only for the last round?

If the main purpose for the customer is to handle each round as a separate result, we may need to export it on each rounds

src/kili/llm/services/export/dynamic.py
```diff
@@ -173,6 +179,9 @@ def _build_rounds(self, chat_items, annotations, json_interface):
                         has_children = True
                         parent_target = chat_item["id"]
 
+                current_round["annotations"] += [
```
baptiste-olivier, line 182: we'll need to do this if we have a pending round when the loop finish (l203, before the ` rounds.append(current_round)`)

src/kili/llm/services/export/dynamic.py
FannyGaudin: Check with @stephnguyenkili that's ok tu put on the last round

src/kili/llm/services/export/dynamic.py
FannyGaudin: Change it thanks

src/kili/llm/services/export/static.py
```diff
@@ -83,6 +83,8 @@ def _format_json_response(
             result[job_name] = job_value["text"]
         else:
             logging.warning(f"Job {job_name} with mlTask {job_config['mlTask']} not supported")
+    if len(result) != 0:
+        return {"asset": result}
```
baptiste-olivier, line 87: To be consistent, this should be `conversation` as well

FannyGaudin: yes ^^

HNicolas (approved):

PR #1828, opened 2024-12-09
feat(LAB-3098): dynamic llm export `completion` level classification

RuellePaul:


src/kili/llm/services/export/dynamic.py
```diff
@@ -72,8 +74,10 @@ def export(
                         "label_type": label["labelType"],
                         "label": {},
                     }
-                    if formatted_response["turn"]:
-                        label_data["label"]["turn"] = formatted_response["turn"]
+                    if formatted_response["response"]:
```
baptiste-olivier, line 77: I find it ambiguous here to use response.

We already have in Kili terminology jsonResponses, and response.

A turn is a prompt, potentially a pre-prompt and multiple completions. If we want to change this term, we need to find a term that is more specific than response, that doesn't overlap with completion

RuellePaul: I agree that it's confusing, but I want to align what the export returns with the job level values ​​set by the client

Currently :

```
type JobLevel = 'conversation' | 'response' | 'completion' | 'user_prompt' | 'system_prompt';
```

I suggest to change the name `response` to `round`, because a round is a set of two completions. Wdyt ?

baptiste-olivier: yes good idea

baptiste-olivier (approved):

PR #1831, opened 2024-12-11
feat: Add list events method

florianlega:


src/kili/event/presentation/client/event.py
```diff
@@ -0,0 +1,71 @@
...
+        # pylint: disable=line-too-long
+        """List events of my project.
+
+        Args:
+            project_id: Identifier of the project.
+            fields: All the fields to request among the possible fields for the events.
+                Defaults is ["createdAt", "event", "id", "organizationId", "payload", "projectId", "userId"].
+            created_at_gte: Filter events created after this date.
+            created_at_lte: Filter events created before this date.
+            user_id: Filter events by user.
```
FannyGaudin, line 40: user id ?

FannyGaudin (approved):

baptiste-olivier (changes requested):
Don't forget to add : 
1. docs, you can check what is done for LLM, something similar should be enough
2. unit test, mocking graphql call with magic mock
3. e2e tests, as today we don't control what events are registered, and it is an important part for our customers (they react on it using webhooks / plugins)

The 3. could be done in a separate ticket once we know better the critical events

baptiste-olivier (approved):

PR #1842, opened 2025-01-14
fix(LAB-3380): Correct vertices calculation for each type of export + creation of units tests

stchabi:


src/kili/services/export/format/coco/__init__.py
```diff
@@ -466,13 +470,30 @@ def _get_shoelace_area(x: List[float], y: List[float]):
 
     return area
 
+def reverse_rotation_vertices(normalized_vertices, rotation_angle):
```
baptiste-olivier, line 473: Can we define this in `AbstractExporter` instead, to avoid re-defining it in every children class ?

src/kili/services/export/format/coco/__init__.py
```diff
@@ -466,13 +470,30 @@ def _get_shoelace_area(x: List[float], y: List[float]):
...
 
 def _get_coco_geometry_from_kili_bpoly(
-    bounding_poly: List[Dict], asset_width: int, asset_height: int
+    bounding_poly: List[Dict], asset_width: int, asset_height: int, rotation_angle: int
 ):
     normalized_vertices = bounding_poly[0]["normalizedVertices"]
-    p_x = [float(vertice["x"]) * asset_width for vertice in normalized_vertices]
-    p_y = [float(vertice["y"]) * asset_height for vertice in normalized_vertices]
+    new_vertices = reverse_rotation_vertices(normalized_vertices, rotation_angle)
+    p_x = [float(vertice["x"]) * asset_width for vertice in new_vertices]
```
baptiste-olivier, line 495: We should have an integer for every p_x and p_y coordinates, as they correspond to pixel coordinates. Rounding the result should be OK

src/kili/services/export/format/base.py
```diff
@@ -50,6 +50,25 @@ class ExportParams(NamedTuple):
     include_sent_back_labels: Optional[bool]
 
 
+def reverse_rotation_vertices(normalized_vertices, rotation_angle) -> List[Dict]:
+    """Allows to retrieve vertices without rotation."""
+    new_vertices = []
```
aurelienlombard, line 55: You put new in the names everywhere, i would prefer something clearer like, vertices_before_rotate or something similar, it would be easier to know on what we work

aurelienlombard (changes requested):

aurelienlombard (approved):

PR #1851, opened 2025-02-03
fix(LAB-3447): correct calculation of vertices in kili format

stchabi:


src/kili/services/export/format/kili/__init__.py
```diff
@@ -2,14 +2,25 @@
...
 from kili.services.types import Job
 
 
+def _clean_json_response(asset: Dict):
+    if asset.get("latestLabel", {}) and asset["latestLabel"].get("jsonResponse", {}):
+        if "ROTATION_JOB" in asset["latestLabel"]["jsonResponse"]:
+            asset["latestLabel"]["jsonResponse"].pop("ROTATION_JOB")
+    if asset.get("labels"):
+        for label in asset["labels"]:
+            if label.get("jsonResponse", {}) and "ROTATION_JOB" in label["jsonResponse"]:
```
HNicolas, line 19: Why are you mixing 2 syntaxes to do the same thing ?
`"key" in asset` or `asset.get("key", {})`
Use `in` to check if a key is in a dict.

src/kili/services/export/format/kili/__init__.py
```diff
@@ -265,31 +292,49 @@ def _scale_normalized_vertices_pdf_annotation(
 
 
 def _scale_normalized_vertices_image_video_annotation(
-    annotation: Dict, asset: Dict, is_label_rotated: bool
+    annotation: Dict,
+    asset: Dict,
+    _is_label_rotated: bool,
+    rotation: int,
+    _normalized_vertices: Optional[bool],
```
HNicolas, line 299: Make `_normalized_vertices` a bool with a default value if you want to make it optional for the caller. The code will be less subject to errors like checking the value is exactly False or an just a falsy value.

HNicolas (changes requested):

src/kili/services/export/format/kili/__init__.py
HNicolas: No I mean
if "key" in asset:
Instead of
fi asset.get("key", {}):

src/kili/services/export/format/kili/__init__.py
HNicolas: I see, still you can update internal methods to have a nicer api.

HNicolas (approved):

src/kili/services/export/format/kili/__init__.py
```diff
@@ -131,11 +142,21 @@ def _scale_label_vertices(self, label: Dict, asset: Dict) -> None:
             else False
         )
 
+        rotation_val = 0
+        if "ROTATION_JOB" in label["jsonResponse"]:
+            rotation_val = label["jsonResponse"]["ROTATION_JOB"]["rotation"]
+
+        normalized_vertices = True
```
baptiste-olivier, line 149: can't you do directly :

```
normalized_vertices = self.normalized_coordinates is False
```

src/kili/services/export/format/kili/__init__.py
```diff
@@ -131,11 +142,21 @@ def _scale_label_vertices(self, label: Dict, asset: Dict) -> None:
...
+        if self.normalized_coordinates is False:
+            normalized_vertices = False
+
         if self.project["inputType"] == "PDF":
             self._scale_json_response_vertices(
                 label["jsonResponse"],
                 asset,
                 is_label_rotated,
+                rotation_val,
```
baptiste-olivier, line 158: I don't think we can have a ROTATION_JOB in Pdf project

src/kili/services/export/format/kili/__init__.py
```diff
@@ -212,7 +239,11 @@ def _scale_all_vertices(object_, width: int, height: int):
 
 
 def _scale_normalized_vertices_pdf_annotation(
-    annotation: Dict, asset: Dict, is_label_rotated: bool = False
+    annotation: Dict,
+    asset: Dict,
+    is_label_rotated: bool = False,
+    _rotation: int = 0,
```
baptiste-olivier, line 245: Do not add property to functions if they are never used

src/kili/services/export/format/kili/__init__.py
```diff
@@ -265,31 +295,49 @@ def _scale_normalized_vertices_pdf_annotation(
 
 
 def _scale_normalized_vertices_image_video_annotation(
-    annotation: Dict, asset: Dict, is_label_rotated: bool
+    annotation: Dict,
+    asset: Dict,
+    _is_label_rotated: bool,
+    rotation: int,
+    _normalized_vertices: bool,
```
baptiste-olivier, line 302: why does this parameter start with an `_` ?

src/kili/services/export/format/kili/__init__.py
```diff
@@ -265,31 +295,49 @@ def _scale_normalized_vertices_pdf_annotation(
 
 
 def _scale_normalized_vertices_image_video_annotation(
-    annotation: Dict, asset: Dict, is_label_rotated: bool
+    annotation: Dict,
+    asset: Dict,
+    _is_label_rotated: bool,
```
baptiste-olivier, line 300: do we still need this one ?

baptiste-olivier (approved):

PR #1864, opened 2025-02-28
fix(LAB-3518): fix export in coco and voc format for videos

stchabi:


src/kili/services/export/format/coco/__init__.py
```diff
@@ -349,8 +350,20 @@ def _get_images_and_annotation_for_videos(
             width, height = get_frame_dimensions(asset)
             frame_ext = Path(asset["jsonContent"][0]).suffix
 
+        elif (
+            isinstance(asset["jsonContent"], str)
+            and asset["jsonContent"] != ""
+            and requests.get(asset["jsonContent"], timeout=100).status_code == 200
+        ):
+            width, height = get_frame_dimensions(asset)
+            response = requests.get(asset["jsonContent"], timeout=100).json()
```
baptiste-olivier, line 359: You do the exact same request 2 times

src/kili/services/export/format/coco/__init__.py
```diff
@@ -349,8 +350,20 @@ def _get_images_and_annotation_for_videos(
...
+            frame_ext = Path(response["0"]).suffix
+
         # video with shouldUseNativeVideo set to True (no frames available)
-        elif Path(asset["content"]).is_file():
+        elif Path(asset["content"]).is_file() or (
+            asset["content"] != ""
+            and requests.get(asset["content"], timeout=100).status_code == 200
```
baptiste-olivier, line 365: same here, you may trigger a request that was already done in the previous check

stchabi: You mean in case jsonContent and content are equal and are both invalid ?

PR #1869, opened 2025-03-20
fix(LAB-3531): do not check is_large_image method for geosat images

BlueGrizzliBear:


src/kili/services/asset_import/image.py
```diff
@@ -79,9 +79,15 @@ def split_asset_by_upload_type(assets: List[AssetLike], is_hosted: bool):
...
             mime_type = get_mime_type(path)
-            is_large_image = ImageDataImporter.get_is_large_image(path)
-            if is_large_image or mime_type in mime_extensions_that_need_post_processing:
+            if mime_type in mime_extensions_that_need_post_processing:
                 async_assets.append(asset)
             else:
-                sync_assets.append(asset)
+                try:
+                    is_large_image = ImageDataImporter.get_is_large_image(path)
+                    if is_large_image or mime_type in mime_extensions_that_need_post_processing:
```
aurelienlombard, line 87: you don't need to check again the mime_type as it has already been done

aurelienlombard (approved):

PR #1871, opened 2025-03-21
fix(LAB-3572): change query timeout and remove retry mechanism to copy_project

BlueGrizzliBear:
This MR is to fix the issues spotted for the copy_project query for large projects : 
- Timeout : currently, the copy_project requests has a timeout of 60 seconds (which seems light for big projects) -> make it longer (10min) to allow big projects to succeed

- Retry : currently, the copy_project requests will retry if they fail : these requests cannot be stopped because they are run asynchronously - and thus should not be retried (the client can retry it manually) -> will remove retry directive on this request

Adds a lign on the method description accordingly.

### DIFF
--- a/src/kili/entrypoints/cli/project/copy.py
+++ b/src/kili/entrypoints/cli/project/copy.py
@@ -52,6 +52,8 @@ def copy_project(
 
     By default, only the json interface, quality settings and project members are copied.
 
+    By default, this method is not configured with a retry mechanism and its timeout is extended to 10 minutes.
+
     If no `title` is provided, the source project title will be used.
     If no description is provided, the description will be set to an empty string.
 


src/kili/core/graphql/graphql_client.py
```diff
@@ -257,8 +257,13 @@ def execute(
         document = query if isinstance(query, DocumentNode) else gql(query)
         variables = self._remove_nullable_inputs(variables) if variables else None
 
+        no_retry = kwargs.pop("no_retry", False)
```
florianlega, line 260: We do not document it ?

src/kili/adapters/kili_api_gateway/project/operations_mixin.py
```diff
@@ -120,5 +120,9 @@ def copy_project(
             }
         }
 
-        result = self.graphql_client.execute(GQL_COPY_PROJECT, variables)
+        # For the copy_project call, we want a much higher timeout, since it can take a long time for large projects.
+        # Instead of 60 seconds, we pass to waiting for 10 minutes.
+        result = self.graphql_client.execute(
+            GQL_COPY_PROJECT, variables, no_retry=True, timeout=600
```
baptiste-olivier, line 126: It may be confusing to have a negative form in a boolean variable name.

Can we set `retry` to False here, and assume that the value is True by default ?

florianlega (approved):

PR #1879, opened 2025-04-02
fix(LAB-3474): deprecation of projectType and creation of fromDemoProject parameter

stchabi:


src/kili/presentation/client/project.py
```diff
@@ -27,6 +26,7 @@
 from kili.use_cases.tag import TagUseCases
 from kili.utils.logcontext import for_all_methods, log_call
 
+from ...exceptions import IncompatibleArgumentsError
```
RuellePaul, line 29: prefer `from kili.exceptions import ...`

RuellePaul (approved):

PR #1880, opened 2025-04-02
feat: aau i can add asset metadata as a keyvalue list that will be

RuellePaul:


HNicolas (approved):

HNicolas (approved):

src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -409,6 +408,117 @@ def generate_variables(batch: Dict) -> Dict:
...
+            AssetFilters(
+                project_id=ProjectId(project_id), asset_id_in=cast(List[AssetId], asset_ids)
+            ),
+            ["id", "jsonMetadata"],
+            QueryOptions(disable_tqdm=True),
+        )
+
+        json_metadatas = []
+        for i, asset in enumerate(assets):
+            current_metadata = asset.get("jsonMetadata", {}) if asset.get("jsonMetadata") else {}
```
josselinbuils, line 452: Why checking if the value is truthy whereas you already provide a default value in the get?

src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -409,6 +408,117 @@ def generate_variables(batch: Dict) -> Dict:
...
```
josselinbuils, line 505: Same here

josselinbuils (approved):

src/kili/entrypoints/mutations/asset/__init__.py
RuellePaul: `asset.jsonMetadata` could be `None`, which results in a fail after :
`None has no attribute update` 

Default value in the get is used only if `asset` doesn't have `jsonMetadata` key

PR #1885, opened 2025-04-15
feat(LAB-3534): add stepIdIn and stepStatusIn filters

BlueGrizzliBear:


RuellePaul:
Here are some highlights of my tests :

<hr/>

```
>>> assets = kili.assets(project_id=project_id, step_status_in=['TO_DO'])

UserWarning: Field status requested : request fields step and stepStatus instead for this project
```

-> we should not pass `status` by default in `fields` if the project is in workflow V2, to avoid a warning if user does not specify fields explicitly

<hr/>

```
>>> assets = kili.assets(project_id=project_id, step_in=['Labeling'])
>>> print(len(assets))
6
>>> assets = kili.assets(project_id=project_id, step_in=['Reviewing 1'])
>>> print(len(assets))
1
>>> assets = kili.assets(project_id=project_id, step_in=['????'])
>>> print(len(assets))
7
```

-> I think we should add a validator on `step_in` to only allow valid values, and raise if the step name is invalid (like for `step_status_in`)
-> Suggestion : rename `step_in` to `step_name_in` ? At first, I thought it was step ids - might be confusing for user 
-> Should not the step name be `Label` and `Review 1` ? (as you're currently refactoring it, I don't know if it's need to be in this MR)

<hr/>


```
>>> assets = kili.assets(project_id=project_id, step_status_in=['DOING'])
>>> print(len(assets))
1

>>> assets = kili.assets(project_id=project_id, status_in=['ONGOING'])
>>> print(len(assets))
UserWarning: Field status requested : request fields step and stepStatus instead for this project
0
```

-> I think it would be better to raise if user request fields forbidden in workflow V2 (and vice-versa) to avoid incorrect results


<hr />

```
>>> assets = kili.assets(project_id=project_id, step_status_in=['SKIPPED'], skipped=True)
>>> print(len(assets))
UserWarning: Filter skipped given : use filter step_status_in with the SKIPPED status instead for this project
1
>>> assets = kili.assets(project_id=project_id, skipped=True)
>>> print(len(assets))
1
```

-> I think this user warning should be present even if the user did not specify `step_status_in`

RuellePaul (approved):

josselinbuils (approved):

PR #1887, opened 2025-04-24
fix(LAB-3592): update annotation vertices that are Scalars

FannyGaudin:


src/kili/adapters/kili_api_gateway/label/common.py
```diff
@@ -40,8 +40,8 @@ def get_annotation_fragment():
         ),
         video_object_detection_annotation_fragment=fragment_builder(
             (
-                "keyAnnotations.annotationValue.vertices.x",
-                "keyAnnotations.annotationValue.vertices.y",
+                "keyAnnotations.annotationValue.vertices",
+                "keyAnnotations.annotationValue.vertices",
```
RuellePaul, line 44: do we need it two times ?

RuellePaul (approved):

src/kili/adapters/kili_api_gateway/label/common.py
FannyGaudin: ah no that's a mistake ^^

PR #1891, opened 2025-04-29
fix(LAB-3532): handle `external_ids` arg for `add_metadata` and `set_metadata`

RuellePaul:
Previously, we could only add/set metadata using asset_id

This MR add ability to use external_ids

```
kili.add_metadata([{'foo': 'bar'}], external_ids=['hd_video.mp4'], project_id="cm9wi5b1b00e3x30w04mh6g5q")
```

### DIFF
--- a/src/kili/entrypoints/mutations/asset/__init__.py
+++ b/src/kili/entrypoints/mutations/asset/__init__.py
@@ -460,7 +460,7 @@ class MutationsAsset(BaseOperationEntrypointMixin):
 
         assets = self.kili_api_gateway.list_assets(
             AssetFilters(
-                project_id=ProjectId(project_id) if project_id else None,
+                project_id=ProjectId(project_id),
                 asset_id_in=cast(List[AssetId], resolved_asset_ids),
             ),
             ["id", "jsonMetadata"],
@@ -477,7 +477,7 @@ class MutationsAsset(BaseOperationEntrypointMixin):
             json_metadatas.append(current_metadata)
 
         return self.update_properties_in_assets(
-            asset_ids=resolved_asset_ids,
+            asset_ids=cast(List[str], resolved_asset_ids),
             json_metadatas=json_metadatas,
         )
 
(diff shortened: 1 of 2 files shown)


src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -485,16 +507,34 @@ def set_metadata(
...
         """
         if is_empty_list_with_warning("set_metadata", "json_metadata", json_metadata):
             return []
 
+        if (asset_ids is not None and external_ids is not None) or (
+            asset_ids is None and external_ids is None
+        ):
+            raise MissingArgumentError("Please provide either `asset_ids` or `external_ids`.")
+
+        resolved_asset_ids = self._resolve_asset_ids(asset_ids, external_ids, project_id)
```
HNicolas, line 532: My guess is that this method calls the API to retrieve assets by external_ids, if it is the case maybe you can instead retrieve assets on line 534 by external_ids to avoid doing a request just for this.

src/kili/entrypoints/mutations/asset/__init__.py
```diff
@@ -412,16 +412,18 @@ def generate_variables(batch: Dict) -> Dict:
     def add_metadata(
         self,
         json_metadata: List[Dict[str, Union[str, int, float]]],
-        asset_ids: List[str],
         project_id: str,
+        asset_ids: Optional[List[str]] = None,
```
josselinbuils, line 416: Isn't a breaking change to move this argument?

RuellePaul: It's not :
- `add_metadata` (and `set_metadata`) are new functions, not used yet
- Python requires that optional parameters (with default values) must come after all required parameters (without default values).

src/kili/entrypoints/mutations/asset/__init__.py
RuellePaul: Nice catch 👍🏻 I updated the code

josselinbuils (approved):

PR #1895, opened 2025-05-15
feat(LAB-3634): refactor shapefile import using project json_interface

RuellePaul:
- Added support for correctly handling different job types (marker, polyline, rectangle, polygon, semantic)

The previous implementation only correctly handled `semantic` job type. This MR adds proper support for building correct `json_response` for all job type, using project `json_interface`

- Fixed issue with duplicate coordinates in QGIS-exported shapefiles

_Shapefiles from QGIS often "close" annotations, resulting in duplicate coordinates (e.g., first and last points are identical in polygons). This MR removes these duplicate points to ensure valid annotations._

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
 ]
 dependencies = [
   "pandas >= 1.0.0, < 3.0.0",
-  "click >= 8.0.0, < 9.0.0",
+  "click >= 8.0.0, < 8.1.8",
   "requests >= 2.0.0, < 3.0.0",
   "tabulate >= 0.9.0, < 0.10.0",
   "tenacity >= 8.0.0, < 9.0.0",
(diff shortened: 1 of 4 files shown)


aurelienlombard (approved):

pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
 ]
 dependencies = [
   "pandas >= 1.0.0, < 3.0.0",
-  "click >= 8.0.0, < 9.0.0",
+  "click >= 8.0.0, < 8.1.8",
```
aurelienlombard, line 34: Why did you need to modify this ?

aurelienlombard (commented):
There's also the markdown lint pipeline that fails, but it does not seem to be linked to your MR, did you check other MRs ?

baptiste-olivier (approved):

pyproject.toml
RuellePaul: To pass the tests
https://kili-technology.slack.com/archives/C041CGM7YBV/p1747382633530959

aurelienlombard: I missed the message on slack. Ok for me.

PR #1914, opened 2025-06-16
[lts] fix(LAB-3755): add width and height in patch_label_json_respons…

FannyGaudin:
…e for correctly interpolate bbox between keyframes


tests/integration/adapters/kili_api_gateway/test_label.py
```diff
@@ -207,31 +208,50 @@ def test_given_kili_gateway_when_adding_labels_by_batch_then_it_calls_proper_res
...
 
 def test_given_project_with_new_annotations_when_calling_list_labels_it_converts_to_json_response(
-    graphql_client: GraphQLClient, http_client: HttpClient
+    graphql_client: GraphQLClient, http_client: HttpClient, mocker: pytest_mock.MockerFixture
 ):
     # Given
-    def mocked_graphql_execute(query, variables, **kwargs):
-        if "projects(" in query:
-            return {"data": [{"inputType": "VIDEO", "jsonInterface": test_case_1.json_interface}]}
+    mocker.patch("pathlib.Path.is_file", return_value=True)
```
FannyGaudin, line 214: I needed to mock Path.is_file(), otherwise the test was failing, but I still don't understand why because the asset have the content properties which is a file.

PR #2039, opened 2026-04-01
fix(LAB-4295): add differents methods to handle review and labeling steps

stchabi:


src/kili/presentation/client/project_workflow.py
```diff
@@ -144,3 +144,136 @@ def copy_workflow_from_project(
             source_project_id=ProjectId(source_project_id),
             destination_project_id=ProjectId(destination_project_id),
         )
+
+    @typechecked
+    def add_review_step(
+        self,
+        project_id: str,
+        name: str,
+        assignees: list[str],
```
RuellePaul, line 153: [product] for `assignees`, shouldn't we use user email insteaf of ids? or have both ?

src/kili/presentation/client/project_workflow.py
```diff
@@ -144,3 +144,136 @@ def copy_workflow_from_project(
...
+        step_coverage: int | None = None,
+        use_honeypot: bool | None = None,
+        send_back_to_step: str | None = None,
```
RuellePaul, line 156: [product] same here, should we use step name instead of step id for better UX ?

src/kili/presentation/client/project_workflow.py
```diff
@@ -144,3 +144,136 @@ def copy_workflow_from_project(
...
             destination_project_id=ProjectId(destination_project_id),
         )
+
+    @typechecked
+    def add_review_step(
+        self,
+        project_id: str,
+        name: str,
+        assignees: list[str],
+        step_coverage: int | None = None,
```
RuellePaul, line 154: Image

we should validate this field in the backend at least (maybe in SDK too), i was able to put forbidden values (negative and above 100)

Image

src/kili/presentation/client/project_workflow.py
RuellePaul: (this comment applies for all new methods)

src/kili/presentation/client/project_workflow.py
RuellePaul: (this comment applies for all new methods)

src/kili/presentation/client/project_workflow.py
```diff
@@ -144,3 +144,136 @@ def copy_workflow_from_project(
...
+        return ProjectWorkflowUseCases(self.kili_api_gateway).update_review_step_properties(
+            project_id=project_id,
+            step_id=step_id,
+            step_coverage=step_coverage,
+            send_back_to_step=send_back_to_step,
+            use_honeypot=use_honeypot,
+        )
+
+    @typechecked
+    def delete_step(
```
RuellePaul: when I tried this method, I got the error :

`typeguard.TypeCheckError: the return value (list) is not a dict`

it seems that delete steps returns a list

RuellePaul: [product] when trying to delete a step in the middle of the workflow, I got the error :

`There must be continuity between step orders (missing 6)`

It seems we cannot delete steps in the middle of the workflow - shouldn't we have a method "delete_last_step" ?

florianlega (approved):

src/kili/use_cases/project_workflow/__init__.py
```diff
@@ -289,6 +294,114 @@ def _remap_send_back_step_ids(
...
+        self,
+        project_id: str,
+        name: str,
+        assignees: list[str],
+        step_coverage: int | None = None,
+        use_honeypot: bool | None = None,
+        send_back_to_step: str | None = None,
+    ) -> dict[str, object]:
+        """Add a review step to a project workflow."""
+        if step_coverage and (step_coverage < 0 or step_coverage > 100):
```
RuellePaul: in python, we can simplify to :
```if step_coverage and not (0 <= step_coverage <= 100):```

File: src/kili/use_cases/project_workflow/__init__.py [hunk 6]
```diff
@@ -289,6 +294,114 @@ def _remap_send_back_step_ids(
...
+        use_honeypot: bool | None = None,
+    ) -> dict[str, object]:
+        """Update properties of a labeling step."""
+        if consensus_coverage and (consensus_coverage < 0 or consensus_coverage > 100):
+            raise ValueError(
+                "The parameter consensus_coverage must be between 0 and 100 (included)."
+            )
+        steps = self._kili_api_gateway.get_steps(
+            project_id=ProjectId(project_id), fields=["steps.id", "steps.name"]
+        )
```
RuellePaul, line 334: it seems the method on the backend already accept name as id :

```
  """
  The step ID or name to identify the labeling step.
  """
  stepId: String!
```

I think we can remove this step of getting the step name and directly pass the name as an id

RuellePaul: (apply everywhere)

src/kili/adapters/kili_api_gateway/project_workflow/operations_mixin.py
```diff
@@ -177,3 +196,65 @@ def remove_reviewers_from_step(
...
+                assignees_to_add.append(user_id)
+                assignees_added.append(email)
+            else:
+                assignees_not_added.append(email)
+        if assignees_not_added:
+            warnings.warn(
+                "These emails were not added (not found or can not review): "
+                + ", ".join(assignees_not_added)
+            )
+        data.assignees = assignees_to_add
```
RuellePaul: maybe it would be cleaner to use existing method (like `list_activated_project_users`) instead of doing the graphQL query from scratch

src/kili/adapters/kili_api_gateway/project_workflow/operations_mixin.py
```diff
@@ -177,3 +196,65 @@ def remove_reviewers_from_step(
             )
 
         return removed_emails
+
+    def add_review_step(self, data: AddReviewStepInput) -> dict:
```
RuellePaul, line 201: I've got two strange behaviors during my tests :

1)
- add a review step "Test"
- remove this step in the UI
- add a review step "Test" -> got an error (step already exists), but step is here again

-> we should not raise an error when reactivating a step IMO

2)
- add a review step "Test"
- remove this step in the UI
- add a review step "Test 2" -> Test is present again

src/kili/presentation/client/project_workflow.py
```diff
@@ -144,3 +144,133 @@ def copy_workflow_from_project(
...
+        )
+
+    @typechecked
+    def update_review_step_properties(
+        self,
+        project_id: str,
+        step_name: str,
+        step_coverage: int | None = None,
+        send_back_to_step: str | None = None,
+        use_honeypot: bool | None = None,
```
RuellePaul, line 217: we can't update assignees of the review step here ?

src/kili/use_cases/project_workflow/__init__.py
```diff
@@ -289,6 +294,114 @@ def _remap_send_back_step_ids(
             )
         return {}
 
+    def add_review_step(
+        self,
+        project_id: str,
+        name: str,
```
RuellePaul: should be `step_name` for harmonization with other methods

src/kili/adapters/kili_api_gateway/project_workflow/operations_mixin.py
```diff
@@ -177,3 +196,65 @@ def remove_reviewers_from_step(
...
+            warnings.warn(
+                "These emails were not added (not found or can not review): "
+                + ", ".join(assignees_not_added)
+            )
+        data.assignees = assignees_to_add
+        variables = {"input": add_review_step_input_mapper(data)}
+        mutation = get_add_review_step_mutation()
+        result = self.graphql_client.execute(mutation, variables)
+        steps = result.get("data", {}).get("steps", [])
+        return next(step for step in steps if step.get("name") == data.name)
```
RuellePaul: `next` in python raise an `StopIteration` error if element is not found - maybe we should add a `None` as the second parameter, and error handling : 

instead of :

```
return next(step for step in steps if step.get("name") == data.name)
```

we should do :

```
step = next(step for step in steps if step.get("name") == data.name, None)
if step is None :  
    raise ...
return step
```

RuellePaul: (apply everywhere when pertinent)

RuellePaul (approved):

PR #2045, opened 2026-05-12
feat(lab-4341): check file type before importing

florianlega:


src/kili/services/asset_import/constants.py
```diff
@@ -10,3 +10,29 @@
...
+            ".jp2",
+            ".ntf",
+            ".nitf",
+        }
+    ),
+    "LLM_RLHF": frozenset({".json"}),
+    "PDF": frozenset({".pdf"}),
+    "TEXT": frozenset({".txt", ".csv"}),
+    "VIDEO": frozenset({".mp4", ".mkv", ".3gp", ".avi", ".m4v", ".mov", ".webm"}),
+    "VIDEO_LEGACY": frozenset({".mp4", ".mkv", ".3gp", ".avi", ".m4v", ".mov", ".webm"}),
```
FannyGaudin: What is this type VIDEO_LEGACY needed for ?

FannyGaudin (approved):

src/kili/services/asset_import/constants.py
florianlega: It's indeed not used anymore by the backend, i'll remove it

src/kili/services/asset_import/__init__.py
```diff
@@ -35,6 +35,8 @@
     "LLM_RLHF": LLMDataImporter,
 }
 
+check_file_extensions = BaseAbstractAssetImporter.check_file_extensions
```
aurelienlombard: [SUGGESTION] Unused module-level alias

`check_file_extensions` is assigned here but never referenced — the actual call on line 68 uses `asset_importer.check_file_extensions(...)` (the instance/static method directly). Consider removing this alias unless you intend it as a convenience export for other modules or external callers.

src/kili/services/asset_import/base.py
```diff
@@ -524,6 +526,43 @@ def check_asset_contents(assets: list[AssetLike]) -> None:
...
+
+        for asset in assets:
+            content = asset.get("content")
+            if isinstance(content, str):
+                ext = get_ext(content)
+                if ext and ext not in allowed:
+                    raise ImportValidationError(
+                        f"File extension '{ext}' is not allowed for {input_type} projects."
+                        f" Allowed extensions: {', '.join(sorted(allowed))}"
+                    )
```
aurelienlombard, line 553: [SUGGESTION] Include asset identifier in error message

When importing many assets, this error doesn't tell the user *which* asset failed. Consider including the external_id to make debugging easier:

```python
raise ImportValidationError(
    f"File extension '{ext}' is not allowed for {input_type} projects"
    f" (asset external_id='{asset.get('external_id', 'unknown')}')."
    f" Allowed extensions: {', '.join(sorted(allowed))}"
)
```

Same applies to the multi-layer error below (line 561-564).

src/kili/services/asset_import/constants.py
```diff
@@ -10,3 +10,28 @@
...
+            ".webp",
+            ".ico",
+            ".tif",
+            ".tiff",
+            ".jp2",
+            ".ntf",
+            ".nitf",
+        }
+    ),
+    "LLM_RLHF": frozenset({".json"}),
```
aurelienlombard, line 33: [QUESTION] TEXT extensions completeness

Only `.txt` and `.csv` are allowed for TEXT projects. Does this match the frontend's `fileExtensionsState` exactly? Other text-like formats (`.html`, `.md`, `.xml`) might be relevant depending on how TEXT projects are used. If the frontend also restricts to just these two, this is fine.

tests/unit/services/asset_import/test_import_common.py
```diff
@@ -24,7 +24,8 @@ def test_cannot_upload_an_image_to_video_project(self, *_):
         url = "https://storage.googleapis.com/label-public-staging/car/car_1.jpg"
         path_image = self.downloader(url)
         assets = [{"content": path_image, "external_id": "image"}]
-        with pytest.raises(MimeTypeError):
+        # Extension check runs before MIME type check, so ImportValidationError is raised first
+        with pytest.raises(ImportValidationError):
```
aurelienlombard, line 28: [QUESTION] Behavioral change: different exception type

Previously, uploading a `.jpg` to a VIDEO project raised `MimeTypeError` (from the MIME type check). Now it raises `ImportValidationError` (extension check runs first). Both exception classes inherit directly from `Exception` — they aren't related by inheritance. Any caller catching `MimeTypeError` specifically would miss this new error.

Since these are internal exceptions (not public SDK API), this is likely fine, but worth confirming no downstream code relies on catching `MimeTypeError` for this scenario.

tests/unit/services/asset_import/test_import_common.py
```diff
@@ -84,3 +85,184 @@ def test_import_assets_verify(self, mocked_verify_batch_imported, *_):
         mocked_verify_batch_imported.assert_not_called()
         import_assets(self.kili, ProjectId("project_id"), assets, verify=True)
         mocked_verify_batch_imported.assert_called_once()
+
+
+@patch("kili.utils.bucket.generate_unique_id", mocked_unique_id)
+@patch("kili.utils.bucket.request_signed_urls", mocked_request_signed_urls)
```
aurelienlombard, line 91: [PRAISE] Thorough test coverage

All input types covered with both rejection and acceptance cases, plus edge cases like extensionless URLs, raw text content, and multi-layer content validation. Well structured.

aurelienlombard (commented):
## Auto Review Summary

**Verdict:** Ready to merge

| Severity | Count |
|----------|-------|
| Blocker | 0 |
| Required | 0 |
| Suggestion | 2 |
| Question | 2 |
| Praise | 1 |

Clean implementation of file extension validation in the import service. The validation logic is well-placed, correctly handles URLs (stripping query params), local paths, multi-layer content, and gracefully skips extensionless content. Tests are thorough.

**Ticket context (LAB-4341):** The ticket asks for SDK-side extension validation before calling the import service, matching the frontend behavior. The "Acceptance criteria" section is empty — consider filling it in for traceability.

src/kili/services/asset_import/constants.py
aurelienlombard: Not relevant i think, as for html we use rich text and use for this the jsonContent

tests/unit/services/asset_import/test_import_common.py
aurelienlombard: I  don't think people were using this before

aurelienlombard (approved):

PR #2046, opened 2026-05-13
fix(LAB-4399): fix to ensure if there is already an extension to not add a second one

stchabi:


src/kili/use_cases/asset/media_downloader.py
```diff
@@ -209,8 +209,9 @@ def get_download_path(
 ) -> Path:
     """Build the path to download a file the file in local."""
     extension = get_file_extension_from_headers(url, http_client)
-    filename = external_id
-    if extension is not None and not filename.endswith(extension):
+    filename = str(external_id)
+    current_suffix = Path(filename).suffix.lower()
+    if extension is not None and current_suffix == "":
```
aurelienlombard: [QUESTION] Behavior change for external IDs with non-extension dots

The old logic (`not filename.endswith(extension)`) would still append the correct extension when the filename had dots that weren't real extensions — e.g., `"2024.01.15_report"` would become `"2024.01.15_report.pdf"`. The new logic skips extension appending whenever *any* suffix exists, so that same filename would stay as `"2024.01.15_report"` (no extension), making it harder to open.

Is this an acceptable tradeoff? If not, a more robust approach would check whether the existing suffix is a known MIME type:

```python
import mimetypes

current_suffix = Path(filename).suffix.lower()
has_known_extension = current_suffix != "" and mimetypes.types_map.get(current_suffix) is not None
if extension is not None and not has_known_extension:
    filename = filename + extension
```

This would preserve both fixes: no double extension for `.jpeg`/`.jpg` mismatches, and proper extension appending for non-extension dots.

src/kili/use_cases/asset/media_downloader.py
```diff
@@ -209,8 +209,9 @@ def get_download_path(
 ) -> Path:
     """Build the path to download a file the file in local."""
     extension = get_file_extension_from_headers(url, http_client)
-    filename = external_id
-    if extension is not None and not filename.endswith(extension):
+    filename = str(external_id)
+    current_suffix = Path(filename).suffix.lower()
```
aurelienlombard, line 213: [SUGGESTION] Unnecessary `.lower()` call

`current_suffix` is only compared against `""` on the next line. Lowercasing an empty string still yields `""`, so `.lower()` has no effect here. It could be removed for clarity:

```python
current_suffix = Path(filename).suffix
```

aurelienlombard (commented):
## Auto Review Summary

**Verdict:** Needs discussion

| Severity | Count |
|----------|-------|
| Blocker | 0 |
| Required | 1 |
| Suggestion | 1 |
| Question | 2 |
| Praise | 1 |

The fix correctly addresses the double-extension bug (e.g., `file.jpeg.jpg`) by switching from `endswith()` to `Path.suffix`, but introduces a subtle behavior change for external IDs containing dots that aren't real file extensions, and is missing unit tests for the fixed scenario.

src/kili/use_cases/asset/media_downloader.py
```diff
@@ -209,8 +209,9 @@ def get_download_path(
 ) -> Path:
     """Build the path to download a file the file in local."""
     extension = get_file_extension_from_headers(url, http_client)
-    filename = external_id
-    if extension is not None and not filename.endswith(extension):
+    filename = str(external_id)
```
aurelienlombard, line 212: [REQUIRED] Missing unit test for the double-extension scenario

This is a bug fix but no test was added to cover the specific case it fixes (e.g., `externalId="file.jpeg"` with a content-type header that resolves to `.jpg`). The existing tests in `test_media_downloader.py` only cover cases where the extension matches (`car_1.jpg` → `.jpg`). A targeted unit test for `get_download_path` that mocks `get_file_extension_from_headers` would prevent regressions:

```python
def test_get_download_path_no_double_extension(mocker):
    mocker.patch(
        "kili.use_cases.asset.media_downloader.get_file_extension_from_headers",
        return_value=".jpg",
    )
    http_client = mocker.MagicMock(spec=HttpClient)
    path = get_download_path("https://example.com/img", "photo.jpeg", Path("/tmp"), http_client)
    assert path.name == "photo.jpeg"  # not "photo.jpeg.jpg"
```

aurelienlombard, line 212: [QUESTION] Missing acceptance criteria on LAB-4399

The Linear ticket has empty "Why", "Acceptance criteria", and "How" sections. Could you fill in the acceptance criteria so reviewers can verify this PR covers all expected scenarios? For example: what happens with `.jpeg`/`.jpg` mismatches, uppercase extensions, external IDs with dots that aren't extensions, etc.

aurelienlombard, line 212: [PRAISE] Good defensive `str()` cast

`AssetExternalId` is a `NewType` of `str`, but callers pass `asset["externalId"]` from an untyped `Dict`, so this explicit `str()` cast is a sensible safety measure.

src/kili/use_cases/asset/media_downloader.py
aurelienlombard: would go for claude suggestion

src/kili/use_cases/asset/media_downloader.py
aurelienlombard: yes agree that it could be nice to add unit tests to be sure we don't remove this later

src/kili/use_cases/asset/media_downloader.py
stchabi: With the modification above the .lower() call is now needed

aurelienlombard (approved):

HNicolas (approved):

PR #2064, opened 2026-08-20
feat(LAB-4623): add a SDK utility to repair PDF unicode maps

aurelienlombard:
Closes [LAB-4623](https://linear.app/kili-technology/issue/LAB-4623/add-a-sdk-utility-to-fix-pdfs-unicode-maps) · relates to SUP-2799, SUP-2794

## Why

Some PDFs describe their own characters incorrectly. The page renders perfectly, but the invisible text layer underneath records the wrong characters — so annotations capture the wrong text, and in some cases cannot be saved at all.

| What the page shows | What the text layer says |
| --- | --- |
| `offset −3.5 mm` | `offset  3.5 mm` — the minus sign is missing |
| `pressure ≥ 20 bar` | `pressure $ 20 bar` — a dollar sign instead |

The second is the dangerous one: nothing looks broken, so the mistake reaches the labelled data unnoticed. This is a fault in the software that produced the document, so it is not the application's job to fix — but customers need a way to deal with it.

## What this adds

`repair_pdf(input_dir, output_dir)` reads every PDF in a directory, corrects the character maps, and writes the result to an output directory under the same file name. Documents needing no correction are copied through, so the output is always a complete set ready to import. A report is printed and returned.

```python
from kili.client import Kili

kili = Kili()
report = kili.repair_pdf(input_dir="documents", output_dir="documents_repaired")
```

Available on both APIs:

| | |
| --- | --- |
| classic client | `kili.repair_pdf(...)` |
| domain API | `kili.assets.repair_pdf(...)` |
| direct | `kili.utils.pdf.repair_pdf(...)` |

**Marked beta**, with a warning raised at call time telling customers to report documents that are not repaired — or that the repair makes worse — so the reference tables can be extended.

## How it works

Repairs are written as **incremental updates**: the original bytes are preserved, page content is untouched, and the document renders exactly as before. Only the invisible text layer changes. Input files are never modified.

Characters are resolved from information already inside the file — glyph names, published character tables (Adobe Glyph List plus the TeX extensions), the PDF base encodings, and a small set of tables for fonts known to describe themselves incorrectly. **Where nothing justifies a character, it is left alone** rather than replaced with a guess: writing a plausible wrong character is worse than writing none, because it is invisible downstream.

Two tables of known problem fonts ship with it — TeX Computer Modern maths symbols, matched on a fingerprint of the whole font rather than its name, and the symbol fonts emitted by Arbortext Advanced Print Publisher, whose glyph names do not describe their glyphs.

## Report

```
PDFs found                          12
Repaired                             2
Already correct                     10
Wrong characters put right          14
Missing characters filled in       640

  file                                          put right  filled in   fonts
  specification.pdf                                    12        640      18
  appendix.pdf                                          2          0       1
```

"Put right" counts characters the document described **wrongly** — the ones that would have corrupted annotations. "Filled in" counts characters it did not describe at all. Both are per character definition, not per occurrence.

## Dependencies

`pypdf` and `fonttools` are optional, following the existing pattern for `kili[cli]` / `kili[image]`:

```bash
pip install kili[pdf]
```

Nothing is imported on the `import kili` path; a clear `ImportError` points at the extra if it is missing.

## Testing

**23 unit tests**, each building its PDF from scratch so it states exactly which defect it covers — no sample documents are committed.

Validated end to end against **286 documents** spanning ~20 publishing tools (pdfTeX, dvips+Ghostscript, Chrome/Skia, PDFlib, Distiller+iText, InDesign, Word, Quartz, xml2rfc) and **23 scripts**:

| corpus | files | repaired | characters corrected | made worse |
| --- | --- | --- | --- | --- |
| Public (arXiv, PLOS, Nature, BMC, institutional) | 175 | 115 | 536 | 0 |
| Kili `PDF_medic` demo project | 20 | 16 | 13 | 0 |
| Mixed local documents | 91 | 26 | 1323 | 0 |

Checks that matter as much as the counts:

- **No document was made worse.** Each run is compared character-class by character-class against the original text, flagging any file where letters disappear without being replaced by the symbol they really were.
- **Documents in non-Latin scripts are left untouched** — all 30 multi-script documents (Arabic, Hebrew, CJK, Thai, Indic, Cyrillic, Greek …) come back byte-identical.
- **No half-characters written** across all 286 repaired documents.
- Page content streams are byte-identical after repair; repairing an already-repaired file is a no-op.

## Known limitations

Documented on the docs page rather than hidden:

- Documents with **no text layer at all** (a scan) cannot be repaired.
- Some **mathematical and decorative symbols** stay unresolved where the document describes them nowhere.
- Documents in **scripts other than the Latin alphabet** are left untouched.

The docs also set out the alternatives — an OCR tool, or an LLM model with vision capabilities — with the trade-offs of each, so customers can choose knowingly.

## Documentation

New page `docs/sdk/pdf_repair.md`, added to the nav: why it happens, the three ways to deal with it, installation, usage on both clients, how to read the report, and what it can and cannot fix.


aurelienlombard:
Closing this as we decided to not put this in the SDK

PR #2071, opened 2026-09-21
feat(LAB-4674): report the assets an assign could not apply to

BlueGrizzliBear:
## Why

`assign_assets_to_labelers` returned the project id once per batch of 100 — the same dict repeated, nothing about the assets — so a caller could not tell which assets the request applied to. Its docstring described that as "a list of dictionaries with the asset ids".

## What

The method reports what the assign did, per asset.

- **Returns `{succeeded, declined, failed}`** instead of a list of `{"id": <project id>}`, merged across the batches it makes so the result is about the assets rather than about how they were chunked.
- **An asset the assign could not be applied to comes back under `declined`**, rather than being absent from a result that said nothing either way. A user at work on an asset — holding its lock **and** having a label on its current step — can no longer be taken off it:
  - calling with an empty user list leaves them assigned
  - calling with another user adds that user, without removing them

  Either way the asset comes back under `declined` rather than `succeeded`.

## Important context

- **Breaking.** Callers reading the old list get a dict. Accepted by Product and absorbed on the support side — the old shape carried no asset information to lose.
- Depends on the backend `assignAssets` mutation (kili!14510), which is not merged yet. This cannot ship before it.
- `declined` carries no reason: no surface reports one, so the SDK does not invent one. `failed` keeps `details`, since a write that threw is worth retrying as is.
- `kili.assets.assign` inherits the new shape — both clients funnel through this one method.
- The backend keeps `assignAssetsToLabelers`, deprecated, so released SDK versions keep working against a new server.

## Validation

Against a backend serving `assignAssets`, on a Workflow V2 project whose labeling step is assigned to `user 1`:

- with `user 1` holding asset 1 open and having labeled it on the current step
  - `kili.assets.assign(asset_ids=[asset_1], to_be_labeled_by_array=[[]])` → `declined` names it, `succeeded` is empty, `user 1` stays assigned
  - `kili.assets.assign(asset_ids=[asset_1], to_be_labeled_by_array=[[user_2]])` → `declined` names it, and the asset ends up assigned to both
- on an asset nobody has open
  - the first call → `succeeded` names it, `declined` is empty


src/kili/domain/asset/asset.py
```diff
@@ -75,6 +75,33 @@ class AssetFilters:
...
+
+class AssignAssetsOutcome(TypedDict):
+    """What an assign did. Every asset given comes back under exactly one key.
+
+    Why an asset was declined is not carried: the reasons read as noise next to the count, so
+    every surface that reports these counts them without saying why.
+    """
+
+    succeeded: list[str]
+    declined: list[AssetRef]
```
josselinbuils, line 101: Why don't we use the same format for succeed on declined assets?

josselinbuils (approved):

src/kili/domain/asset/asset.py
BlueGrizzliBear: Addressed alongside https://gitlab.com/kili-technology/kili/-/merge_requests/14510#note_3890126002

PR #2073, opened 2026-09-22
fix(LAB-4426): honour normalized_coordinates on a pixel-labeled export

RuellePaul:
Enabled Intelligence raised four points against the SDK while running **25.2.10**, which
carries none of LAB-4426 — the first release that does is **26.1.12**, published
2026-09-17. Two are answered by that upgrade alone. This PR carries the one real SDK
defect among them, plus a docs fix.

The label-creation half that was here initially has been dropped after @aurelienlombard's
review: it belongs in the backend, not here. See *What is not here* below.

Verified against a live `LTS-2026-1-experimental` stack (kili, import, export,
dynamic-tiler all on that branch) with a real MSP 1.11.0 instance and the sample NITF —
2948x2960, RPC metadata, no usable geotransform.

## `normalized_coordinates` was ignored on a pixel-labeled export

The option is read on every other export path and was dropped on this one: whatever was
passed, the Kili export of a pixel-labeled geospatial project came out with the pixel
coordinates beside the fractions. Reproduced against published 26.1.12 —
`normalized_coordinates=True` still returned `vertices: {x: 737.0, y: 740.0}`.

`True` now leaves the fractions alone. `None` and `False` keep adding the pixel
coordinates, which is where this deliberately differs from an image project, whose default
is normalized-only: on this mode the absolute values are the point, and a consumer that
never passes the option must keep receiving them.

Measured end to end on the imported asset, a box at 0.25/0.25 on a 2948x2960 image:

| `normalized_coordinates` | `normalizedVertices` | `vertices` |
|---|---|---|
| `None` | `{0.25, 0.25}` | `{737.0, 740.0}` |
| `False` | `{0.25, 0.25}` | `{737.0, 740.0}` |
| `True` | `{0.25, 0.25}` | absent |

Worth passing on to the client: they reported calling
`kili.export_labels(normalized_vertices=False)`. There is no such parameter — it is
`normalized_coordinates`, and `normalized_vertices=...` raises `TypeError` before anything
is exported.

## Where a geospatial asset's dimensions live — docs only

`kili.assets(fields=["resolution.width", "resolution.height"])` is the documented way to
get an asset's size, and on a geospatial asset it returns `None` — which reads as the
dimensions not being available at all. They are carried per layer by
`geospatialExportMetadata`, alongside `mspIsd` and the layer's CRS. Confirmed on the
imported asset: `resolution` -> `None`, `geospatialExportMetadata[0]` -> `width 2948,
height 2960, labelingCRS PIXEL`.

## What is not here

**Creating a label on a CRS-locked geospatial project still fails**, with
`[failedLabelCreation] CRS code is required for geospatial projects with a locked CRS`.
This PR first carried an SDK-side fix that derived the CRS from
`project.geospatialSettings.labelingCRSCode`. The review was right to block it, and it is
gone:

- the server already holds that same string one line above the call
  (`projectLockedCRSCode`, `services/label/mutations.ts:879`), so the
  `asset.labelingCRSCode !== crsCode` check could never catch a real mistake;
- every other part of the system treats the **asset** as the source of truth once pinned
  (`useLeafletCRS.ts:110`, *"so that project-level CRS changes don't affect already-annotated
  assets"*), so the guessed value is wrong wherever asset and project disagree;
- `'original'` is a valid project `labelingCRSCode` (`LABELING_CRS_CODES`) and makes
  `projectHasLockedCrs` truthy, so an SDK-first label would pin a non-CRS sentinel onto
  `asset.labelingCRSCode` — a column never reset — leaving the labeler stuck on
  `LabelReadOnlyReason.CrsLocked`.

The requirement is over-broad rather than the SDK being at fault: the labeling interface's
own `pages/project/queue/hooks/useAppendToLabelsFromBatch.ts` sends no `crsCode` either and
fails the same way, while `redux/asset-label/helpers/appendToLabels.ts:218` sends the CRS
the labeler actually had on screen. A backend MR
scoping the requirement to the labeling interface follows; that, not this PR, is what
unblocks the client.

## Tests

`795 passed, 1 skipped` on the CI's own pytest invocation, coverage 75.81% (gate 75%),
pylint 10.00/10 and pyright clean on `src/kili`. Three new unit tests pin the export's
behaviour on each value of the option.

https://claude.ai/code/session_0178uQWPtkFEWyAH7r4FDX5d


src/kili/use_cases/label/__init__.py
```diff
@@ -47,6 +47,24 @@
...
+        if project_id is None:
+            return None
+
+        project = self._kili_api_gateway.get_project(
+            project_id=project_id, fields=["inputType", "geospatialSettings"]
+        )
+        if project.get("inputType") != "GEOSPATIAL":
+            return None
+
+        return (project.get("geospatialSettings") or {}).get("labelingCRSCode")
```
aurelienlombard: [BLOCKER] The SDK can't know the CRS — this guesses it from the project

The CRS a label is in is a fact about the labeling session, which is why `appendToLabels.ts:303` sends `leafletCRSCode`: the CRS the labeler actually had on screen. The SDK has no equivalent, so this reads `project.geospatialSettings.labelingCRSCode` and echoes it back — and the server already has that string, one line above the call (`projectLockedCRSCode`, `services/label/mutations.ts:879`). The `asset.labelingCRSCode !== crsCode` check can therefore never catch a real SDK mistake, since both sides derive from the same field.

Worse, the value is wrong wherever the asset and the project disagree — and every other part of the system treats the asset as the source of truth once pinned (`computeDefaultCRSCodeForGeospatialProject` checks `assetLabelingCRSCode` first, *"so that project-level CRS changes don't affect already-annotated assets"*; `useIsReadOnly` uses `assetLabelingCrsCode ?? projectEffectiveCrs`; the backend compares against the asset and never the project):

- **Stale project setting.** Asset labeled in the UI under `EPSG:4326`, admin switches the project to `EPSG:3857`. The UI keeps labeling that asset in 4326, by design. The next SDK label stamps 3857 → `CRS mismatch ... expected "EPSG:4326" (locked on first label) but received "EPSG:3857"`, on an asset the UI still accepts.
- **The `'original'` sentinel.** It's a valid `labelingCRSCode` and `projectHasLockedCrs` is `!!labelingCRSCode`, so this returns the literal string. The frontend explicitly refuses to read it as a CRS (*"'original' is a sentinel meaning 'use the asset's native CRS' — skip it here"*). An SDK-first label pins `'original'` onto `asset.labelingCRSCode` — whose only writer is `mutations.ts:181`, never reset — and the UI then returns it verbatim from the asset branch, leaving the labeler on `LabelReadOnlyReason.CrsLocked`.

The requirement is over-broad rather than the SDK being at fault: the frontend's own `appendToLabelsFromBatch` (`redux/asset-label/actions.ts:716`) sends no `crsCode` either and fails the same way. The exemption in `enforceCrsForGeospatialLabels` was meant for programmatic uploads but is spelled as *prediction and inference*.

Suggest dropping this method and `crs_code` entirely, and fixing it in `enforceCrsForGeospatialLabels`: require a `crsCode` only from the labeling interface — `context.kiliClient?.name === 'label-frontend'` (`index.ts:228-238` populates it from `kili-client-name ?? Apollographql-Client-Name`; `project/mutations.ts:55` already branches on `kiliClient`). Every other caller keeps today's skip-and-don't-pin behaviour, so the UI still pins the first labeler's CRS and nothing writes a guessed one. The header is advisory, which is fine for a consistency pin.

The export half of this PR is independent and stands on its own.

aurelienlombard (commented):
## Auto Review Summary

**Verdict:** Needs changes

| Severity | Count |
|----------|-------|
| Blocker | 1 |

Checked against `LTS-2026-1-experimental`. The export fix is sound. The label fix derives the labeling CRS from the project setting, but the backend, the labeling interface and the read-only check all treat the asset's pinned CRS as the source of truth — so a project whose CRS changed after labeling, or one set to the `'original'` sentinel, gets a mismatch or a non-CRS pinned onto the asset row. The requirement that forces this looks like it should be scoped to the labeling interface instead.

*Related to LAB-4426 (Done); the ticket's SDK criteria shipped in #2058, this is follow-up repair.*

src/kili/use_cases/label/__init__.py
RuellePaul: Agreed on all three points — verified each against `LTS-2026-1-experimental` itself
(`96d6d68c`), and the commit is dropped. The PR is now the export fix and the docstring
only.

What settled it for me is a fourth case you didn't cite: the labeling interface's own
batch path, `pages/project/queue/hooks/useAppendToLabelsFromBatch.ts`, sends no `crsCode`
either, where `redux/asset-label/helpers/appendToLabels.ts:218` does. So the UI already
fails this requirement on one of its two paths — the SDK was never the outlier, and
scoping the requirement rather than teaching every caller to guess is clearly the right
direction.

Two small corrections to the references, for whoever picks up the backend change:
`appendToLabels.ts:303` and `redux/asset-label/actions.ts:716` are not the paths on this
branch — they are the two files above. And the exemption is already spelled as
*"labels without a crsCode (e.g. programmatic API uploads) are skipped"* (`mutations.ts:99`),
not prediction and inference, which only strengthens your reading: programmatic uploads
were meant to be exempt and `projectHasLockedCrs` silently revoked that.

One extra note on `'original'`: `LABELING_CRS_CODES` on this branch also carries
`PIXEL_LABELING_CRS_CODE`, so the sentinel problem and the pixel case share the same
`!!labelingCRSCode` test — worth covering both when the gate moves to
`context.kiliClient?.name`.

I'll take the backend MR.

aurelienlombard (approved):

